PR #5, opened 2016-01-28
VALIDATOR-387 Userinfo without colon should be valid in UrlValidator

flexfrank:


PascalSchumacher:
Thanks for the pull request!

It was merged in https://github.com/apache/commons-validator/commit/ab1f195d44a61b2a01bb4c3f626afcacda978867 and released with commons-validator 1.6.

It would be nice if you could close this pull request

PR #7, opened 2016-08-23
Validating ISIN codes

raupachz:
An International Securities Identification Number (ISIN) uniquely identifies a security. Its structure is defined in ISO 6166. The isin uses country codes and Luhn algorithm for sanity checking.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/ISINValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/ISINValidator.java
@@ -23,7 +23,7 @@ import org.apache.commons.validator.routines.checkdigit.LuhnCheckDigit;
 /**
  * International Securities Identification Number (ISIN) uniquely identifies a
  * bonds, commercial paper, stocks or warrant. The structure is governed
- * by ISO 6166.
+ * by <a href="http://www.anna-web.org/standards/isin-iso-6166/">ISO 6166:2013</a>.
  */
 public class ISINValidator implements Serializable {
     
@@ -45,7 +45,8 @@ public class ISINValidator implements Serializable {
             return false;
         }
 
-        // check country code or international code
+        // check country code or code XS for Euroclear and Clearstream
+        // 
         String countryCode = isin.substring(0, 2);
         if (!countryCode.equals("XS")) {
             boolean foundCode = false;
--- a/src/main/java/org/apache/commons/validator/routines/ISINValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/ISINValidator.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.validator.routines;
+
+import java.io.Serializable;
+import java.util.Locale;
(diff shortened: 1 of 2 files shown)


britter:
Nice! Since this is a bigger contribution I need to ask you to sign an [ICLA](https://www.apache.org/licenses/icla.txt) with the ASF

src/main/java/org/apache/commons/validator/routines/ISINValidator.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.io.Serializable;
+import java.util.Locale;
+import org.apache.commons.validator.routines.checkdigit.LuhnCheckDigit;
+
+/**
+ * International Securities Identification Number (ISIN) uniquely identifies a
+ * bonds, commercial paper, stocks or warrant.
+ */
+public class ISINValidator implements Serializable {
```
britter: There are several base classes in the routines package, which may be of help when implementing this. Have you looked at them? I don't know the validation logic of ISIN, but I could imagine that it has something to do with CheckDigits? Have a look at ISBNValidator and IBANValidator for examples.

src/main/java/org/apache/commons/validator/routines/ISINValidator.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+package org.apache.commons.validator.routines;
+
+import java.io.Serializable;
+import java.util.Locale;
+import org.apache.commons.validator.routines.checkdigit.LuhnCheckDigit;
+
+/**
+ * International Securities Identification Number (ISIN) uniquely identifies a
+ * bonds, commercial paper, stocks or warrant.
```
britter: Maybe add a link to the ISO standard for ISIN?

raupachz:
@britter I think I filed an CLA quite a while ago. Can you check with ASF? I checked the other methods and reuse the Luhn implementation. It is really a simple validation with everything that is already provided by Commons Validation. I will add a link to ISO 6166

src/main/java/org/apache/commons/validator/routines/ISINValidator.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.io.Serializable;
+import java.util.Locale;
+import org.apache.commons.validator.routines.checkdigit.LuhnCheckDigit;
+
+/**
+ * International Securities Identification Number (ISIN) uniquely identifies a
+ * bonds, commercial paper, stocks or warrant. The structure is governed
+ * by ISO 6166.
+ */
```
sebbASF: Is there a link to a public version of the document?

raupachz: Must be this one: http://www.anna-web.org/standards/isin-iso-6166/

src/main/java/org/apache/commons/validator/routines/ISINValidator.java
```diff
@@ -0,0 +1,89 @@
...
+
+        // check invariant
+        if (isin == null || isin.length() != ISIN_LENGTH || !isin.matches("^[A-Z0-9]{12}$")) {
+            return false;
+        }
+
+        // check country code or international code
+        String countryCode = isin.substring(0, 2);
+        if (!countryCode.equals("XS")) {
+            boolean foundCode = false;
```
sebbASF: It's not clear why 'XS' is excluded - need to link to official docs.

niallkp:
The code already exists in Commons Validator to do all this except for the ISO Country Code validation:

- We have an existing ISINCheckDigit routine (so no need to convert the characters and use the LuhnCheckDigit routine)

- We have the CodeValidator which combines regex validation, code length validation & check digit validation. So you can do something like the following:
`CodeValidator validator = new CodeValidator("[A-Z0-9]{12}", 12,ISINCheckDigit.ISIN_CHECK_DIGIT);`

Validation routines are more re-usable when then are split into small dedicated pieces. So I would suggest creating an "ISO Country Code Validator" and then combing that with the above CodeValidator & ISINCheckDigit to compose an "ISINValidator.

Lastly the Locale.getISOCountries() method does an "array copy" every  time it is called. So a new "ISO Country Code Validator" should call that method only once and cache the array to avoid the creation of the array every time.

raupachz:
@niallkp Sorry, my bad. I totally missed the ISINCheckDigit routine. How, did I miss that one? In that case we probably don't need this pull request anymore. The parts are already there. Thank you for the detailed explanation. Keep up the good work!

britter:
@raupachz okay, the let's just close this PR. Thank you!

PR #9, opened 2018-02-16
fix email with dot and whitespace end validation

unintended:


coryfklein:
I just ran into this today. It seems a little absurd that commons-validator performs this way:
```
validator.isValid("andy.noble@data-workshop.com.")  // false
validator.isValid("andy.noble@data-workshop.com. ")  // true
```

Certainly this is a bug? I tried reading over the RFC and it's absolutely greek to me, but it seems that commons-validator should either accept both cases or neither.

coryfklein:
@garydgregory Please forgive me for at-mentioning you, but do you see any merit in reopening this? If it would help, I could create a new PR, or even improve the implementation if you find it lacking.

PR #19, opened 2019-10-20
[VALIDATOR-455] added IBAN validator for Vatican, added also one entry for each missing IBAN type to IBANCheckDigitTest

mikkomaunu:


mikkomaunu:
Errors reported by Travis look unrelated me. For JDK 11 there seems to differences in Date parsing. JDK 13 build fails because source and target should be at least 1.7.

sebbASF:
Implemented

PR #20, opened 2019-11-02
Make sure no end angular bracket like < or > an the bingen and ending…

tdesta7:
This my test of validation


sebbASF:
No idea what this is about.
Please provide a minimal change set.

PR #21, opened 2019-11-02
Added if statement to check if space in email as well as added test a…

RobDOH02:
…round it this is a fix to #449 worked with @malhariri92. This was done as part of the CSCC Haackathon.


sebbASF:
Sorry, but the PR is not accepted for the following reasons:
- spaces *are* allowed in the local part of an email address, provided that they are escaped or part of a quoted string.
- the PR contains a spurious file (.iml)

PR #24, opened 2020-05-26
Fix travis scripts, error in javadoc, and StandardCharsets which cannot be used on java1.6

XenoAmess:


sebbASF:
Please use separate PRs for different fixes

PR #29, opened 2020-06-24
Add Getters for IBANValidator's RegexValidator its patterns

arnaudfnr:
Hi, 

For a project I need to have access to the regex used to validate IBANs, therefore I added a couple of getters to make it possible.

@sebbASF I saw that you have been contributing quite a lot recently, have you planned to perform a release ? If yes, do you have an estimation of its date please ? It would be nice if these changes could be added :pray:

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/RegexValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/RegexValidator.java
@@ -132,6 +132,15 @@ public class RegexValidator implements Serializable {
     }
 
     /**
+     * Getter method for <code>patterns</code>.
+     *
+     * @return the array of <code>patterns</code> used to validate values.
+     */
+    public Pattern[] getPatterns() {
+        return patterns;
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


sebbASF:
Note that the validator field is package-protected so you could get access from the relevant package (as is done by the test class).
Obviously that is not suitable for a project that is publicly available.

src/main/java/org/apache/commons/validator/routines/RegexValidator.java
```diff
@@ -131,6 +131,15 @@ public RegexValidator(String[] regexs, boolean caseSensitive) {
         }
     }
 
+    /**
+     * Getter method for <code>patterns</code>.
+     *
+     * @return the array of <code>patterns</code> used to validate values.
+     */
+    public Pattern[] getPatterns() {
+        return patterns;
```
sebbASF, line 140: The method must return a copy of the the patterns otherwise the caller can mutate the array.

PR #45, opened 2020-10-14
JUnit assertThrows

nhojpatrick:
Code could be cleaner if it could be Java 1.8.

I could set `<maven.compiler.testSource>1.8</maven.compiler.testSource>`... thoughts???

### DIFF
--- a/src/test/java/org/apache/commons/validator/routines/DomainValidatorStartupTest.java
+++ b/src/test/java/org/apache/commons/validator/routines/DomainValidatorStartupTest.java
@@ -16,6 +16,9 @@
  */
 package org.apache.commons.validator.routines;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+import static org.hamcrest.core.IsEqual.equalTo;
 import static org.junit.Assert.*;
 
 import java.util.ArrayList;
@@ -23,6 +26,7 @@ import java.util.List;
 
 import org.apache.commons.validator.routines.DomainValidator.ArrayType;
 import org.junit.Test;
+import org.junit.function.ThrowingRunnable;
 import org.junit.runner.RunWith;
 
 import org.bitstrings.test.junit.runner.ClassLoaderPerTestRunner;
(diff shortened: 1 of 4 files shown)


garydgregory:
I am ok with Java 8 but others might not be. May you please bring this up on the dev ML? It would be better if using Java 8 would bring benefits to the main code.

nhojpatrick:
@garydgregory have continued assertThrows email about maven.compiler.testSource.

About the jdk 1.7 build failure, i'm not sure exactly what the issue is as from what i'm aware i've not touched anything to do with `japicmp/maven/JApiCmpMojo`

src/test/java/org/apache/commons/validator/routines/DomainValidatorStartupTest.java
```diff
@@ -35,24 +39,52 @@
...
     public void testUpdateBaseArrayCC() {
-        DomainValidator.updateTLDOverride(ArrayType.COUNTRY_CODE_RO, new String[]{"com"});
+        // FIXME Simplification once upgraded to Java 1.8
+        final ThrowingRunnable testMethod = new ThrowingRunnable() {
+            public void run() {
+                DomainValidator.updateTLDOverride(ArrayType.COUNTRY_CODE_RO, new String[]{"com"});
+            }
+        };
+        final IllegalArgumentException thrown = assertThrows(IllegalArgumentException.class, testMethod);
+        assertThat(thrown.getMessage(), is(equalTo("Cannot update the table: COUNTRY_CODE_RO")));
```
garydgregory, line 51: Same comment as in the Apache Commons Collections PR. Let's skip asserting the exception messages.

nhojpatrick: Why skip? It's a common conversation I always keep having... I just wondered your rational...

My argument to alway checking the exception message is how do you know the test is failing as you are expecting it to fail?

You write a test to check X, and it throws Y exception. If Z might also causes Y to be thrown, so how do you know your testing X? My view is if you skip checking the exception message then you might as well just delete the test.

So Digester BeanPropertySetterRuleTestCase.testSetUnknownProperty according to the error messages and comments should produce an NoSuchMethodException, but currently is throws org.xml.sax.SAXException due to a code path meaning any non expected Exception is ignored.

So this test is to check COUNTRY_CODE_RO, but if the code behind the test changes and IllegalArgumentException is still thrown, how do you know it's the correct exception for what you have setup?

nhojpatrick:
Could this get another review and try and get it merged.
I'm leaving asserting the exception messages, as it's good practice. You write a test with a set input, the output is also important.

pom.xml
```diff
@@ -132,6 +132,18 @@
     </plugins>
   </build>
 
+  <dependencyManagement>
```
garydgregory, line 135: This is already defined in the parent POM. Is it not working for you?

nhojpatrick: i was rebasing 2-3 year old branches and had not spotted it move into the commons-parent. will correct all the branches i've made.

nhojpatrick:
@garydgregory breaking it out into separate pr's, as i keep needing to rebase and often face merge issues, so easier if 1 pr per test. then it might be easier to get them reviewed and merged quicker.

https://github.com/apache/commons-validator/pull/92
https://github.com/apache/commons-validator/pull/93
https://github.com/apache/commons-validator/pull/94
https://github.com/apache/commons-validator/pull/95

PR #49, opened 2020-11-30
added special prefixes for ISINs

karlkauc:
ISINs can have special prefixes accodring to ISIN ISO Standard.
Missing prefixes:
EU, XA, XB, XC, XD


karlkauc:
I also added two ISO codes as prefix for valid ISINs

garydgregory:
Hello @karlkauc 

Please try git master and see if this PR is still needed. All current cases should be handled by the class now.

garydgregory:
Ping @karlkauc

PR #68, opened 2022-01-07
Add validator for UUIDs

dheid:
Hi @garydgregory !

A lot of times I need to check the validity of incoming parameters and use the Commons Validator library. Thanks for that!

Since I validate UUIDs a lot of times, I wanted to contribute a thoughtful and fast UUID check to Commons Validator. It would be so kind if you could include this.

Thank you very much!

Daniel

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/UUIDValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/UUIDValidator.java
@@ -1,0 +1,371 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


sheelprabhakar (commented):
Looks good to me

dheid:
Very cool @sheelprabhakar ! What are the next steps? Will someone merge this PR?

garydgregory:
I will review this by thelis weekend.

dheid:
@garydgregory So cool, thank you very much!

dheid:
@garydgregory Okay, if your opinion and understanding of tests is a problem and you don't want additional validators, then I'll close this PR.

benrhine:
So we're just going to let this die? Ive been looking for basically this to allow an annotation based validation of UUID parameters or at a minimum a nice concise way to somehow say isValid outside of just a parsing exception. IMHO the reason this is necessary is there are places it would be better to have validation rather than just parse errors.

dheid:
> So we're just going to let this die? Ive been looking for basically this to allow an annotation based validation of UUID parameters or at a minimum a nice concise way to somehow say isValid outside of just a parsing exception. IMHO the reason this is necessary is there are places it would be better to have validation rather than just parse errors.

Hi @benrhine ! The folks from Hibernate Validators were able to read my tests without problems, liked my implementation and merged it into the new version. It will go into version 8.0.0 https://github.com/hibernate/hibernate-validator/pull/1199

dermoritz:
sad about seeing this closed :-(. So to validate a string to be an UUID in Java still needs to copy paste the regex around? pulling hibernate seems to be a bad option

ecki:
From my point offside the test coverage looks fine and while it is a bit unusual it’s easy to read and allows for good error messages. So if nothing else speaks against it, i will merge it.

alexjferreiraml:
> From my point off view the test coverage looks fine and while it is a bit unusual it’s easy to read and allows for good error messages. So if nothing else speaks against it, i will merge it.

So, is it merged?

PR #76, opened 2022-05-20
Updated InetAddressValidator isValid Doc String

GN:
Updated isValid() doc string to clarify that it checks whether or not an address is both an IPv4 or IPv6 address.


garydgregory (changes requested):
No para tags are needed here.

PR #86, opened 2022-10-04
Missing findbugs version

nhojpatrick:
The property `${commons.findbugs.version}` is not defined, so when doing something like `mvn org.apache.maven.plugins:maven-dependency-plugin:3.3.0:go-offline` it will error as it can't resolve the version to use.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,8 @@
 
     <commons.junit.version>5.9.1</commons.junit.version>
     <commons.japicmp.version>0.16.0</commons.japicmp.version>
-  </properties> 
+    <commons.findbugs.version>3.0.5</commons.findbugs.version>
+  </properties>
 
   <scm>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -59,7 +59,8 @@
 
     <commons.junit.version>5.9.1</commons.junit.version>
     <commons.japicmp.version>0.16.0</commons.japicmp.version>
-  </properties> 
+    <commons.findbugs.version>3.0.5</commons.findbugs.version>
```
garydgregory, line 62: -1: Take a look at the rest of the POM and parent POM, FindBugs has been replaced by SpotBugs

nhojpatrick: yep i'm aware findbugs got replace by spotbugs, but you still have findbugs config so i'm fixing that broken config...
i can do a pr to remove findbugs if that is your preference...

PR #100, opened 2022-11-06
CreditCardValidator. Added MIR cards support

apervushin:


garydgregory:
@apervushin ping 🔔 We need a reference to a specification documentation.

PR #115, opened 2023-01-16
add IBAN Validator for Djibouti DJ as listed in iban-registry.pdf ...

homebeaver:
... Release 92 – May 2022

see https://issues.apache.org/jira/browse/VALIDATOR-486

Hi @garydgregory this is the validator for DJ

regards EUGen


garydgregory:
Done in git master with another PR.

PR #129, opened 2023-04-06
ISINValidator. Added additional exception country code

ashoknan2:
Currently ISIN validator missing the country validation for codes like EU. These are exceptional codes. This is detailed in
https://www.iso.org/obp/ui/#iso:pub:PUB500001:en. Added the list of exceptional codes to the validator.


garydgregory:
This PR has no tests.

ashoknan2:
Added the test case

ashoknan2:
Please let me know if this PR has any issues?

PR #130, opened 2023-04-07
Applied refactoring techniques

LalithMaringanti:
Refactoring technique: Push down method 

Class/method:  org.apache.commons.validator.routines.AbstractNumberValidator.determineScale() 

Description: AbstractNumberValidator class has a method named as determineScale() which is used only once in another method of the child class of AbstractNumberValidator which is BigDecimalValidator class. Therefore, removed determineScale() method from the parent class AbstractNumberValidator and placed it in the child class BigDecimalValidator class. This is done to increase cohesiveness and code readability.  

Refactoring technique: Extract class 

Class/Method: org.apache.commons.validator.ValidatorAction  

Description: ValidatorAction consists of methods such as isValid(), handleIndexedField(), getValidationClassInstance() along with methods which are related to validation of only javascript file/functions. For instance readJavaScriptFile(), javaScriptAlreadyLoaded(), formatJavaScriptFileName() etc. Removed all these methods and placed them in a newly created class ValidatorActionUtils to make ValidatorAction more cohesive. 

Refactoring technique: Replace conditional with polymorphism 

Class/Method:  org.apache.commons.validator.ValidatorResources.getParent() 

Description: removed if else conditional statements and created abstract class FormSetHierarchy which consists of abstract method getParent which returns FormSet object. For each if condition, a new class is created which extends the FormSetHierarchy class and implements different getParent method.

Refactoring technique: Extract Method 

Class/Method: org.apache.commons.validator.routines.DomainValidator.unicodeToASCII() 

Description: Extracted all the statements which converts given input string to ascii string and created a new method called toAsciiCode() method and statements responsible for adding trailing dots are moved into addTrailingDotIfMissing() method

### DIFF
--- a/src/main/java/org/apache/commons/validator/CountryFormSet.java
+++ b/src/main/java/org/apache/commons/validator/CountryFormSet.java
@@ -1,0 +1,42 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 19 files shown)


aherbert:
Removing the protected method `protected int determineScale(final NumberFormat format)` from `AbstractNumberValidator` would be a breaking change for binary compatibility. Try this locally:
```
mvn clean package japicmp:cmp
```
You also have many other files in the PR that are unrelated to this change.

garydgregory:
You can also run the default Maven goal by running `mvn` from the command line which will run all checks.

garydgregory:
Closing.

PR #153, opened 2023-09-13
bump commons-commons-collections-4.4

landsman:
Hello,

I noticed in my editor that there is some vulnerability reported.

image

Problem is that affected dependency [commons-collection](https://github.com/apache/commons-collections) do not have a new releases on maven repository: https://repo.maven.apache.org/maven2/commons-collections/commons-collections/

However there are new releases under bigger tags than original 3.2.2, see: 
https://github.com/apache/commons-collections/releases/tag/commons-commons-collections-4.4

I don't know why. Can you please check it with the responsible people? I don't know how to connect to the correct mailing list, write there. I would love to write it as a GitHub issue but it's not possible.

So I did this dump pull request to let you know, and ask you to help me with that.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,7 @@
     <dependency>
       <groupId>commons-collections</groupId>
       <artifactId>commons-collections</artifactId>
-      <version>3.2.2</version>
+      <version>4.4.0</version>
     </dependency>
 
     <!-- Needed for IBANValidatorTest -->


landsman:
Probably @garydgregory will know more about this topic.

garydgregory:
> I tried to bump this dependency locally of course, but I faced this problem with the missing version on the Maven Repository that I mentioned.
> 
> Screenshot 2023-09-13 at 16 06 44
> 
> Yes, I'm aware that's a bad idea, but I don't know a better way to let you know when Issues are not allowed here.

You're not looking in the right place ;-) please see https://commons.apache.org/proper/commons-collections/dependency-info.html

landsman:
Great, that's it. Do you know why it's been renamed? It's kinda unclear to me.

landsman:
Is this dependency needed at all? I did not find any usage in the source code.

garydgregory:
Apache Commons Components normally only break binary compatibility in a new major release, which the justification from going from 3.x to 4.x. When this happens, the Java package name is changed as are the Maven coordinates.

garydgregory:
Taking a closer look, this will not solve your issue: Apache Commons BeanUtils (ACBU) 1.9.4 [depends directly on Apache Commons Collections (ACC) 3.2.2](https://github.com/apache/commons-beanutils/blob/32ceb2c92512d44f97638805e2f3fd9d70dfcfc6/pom.xml#L347-L351). So your change will just add Commons Collections 4.4 to the classpath. You can't remove ACC 3.x because that will break ACBU which will in turn break Validator.

landsman:
You are right. There is promising PR already: https://github.com/apache/commons-beanutils/pull/40
Can you please try to look at it?

garydgregory:
> You are right. There is promising PR already: [apache/commons-beanutils#40](https://github.com/apache/commons-beanutils/pull/40) Can you please try to look at it?

Eventually we will release BeanUtils 2 but it still needs work. Specifically the "2" types must be pulled up into their super types. Feel free to help ;-)

landsman:
Since the problematic dependency is resolved in `commons-beanutils` can you please create a new milestone for release `1.9.5`? We don't know what is needed to release this new version without it.

garydgregory:
> Since the problematic dependency is resolved in `commons-beanutils` can you please create a new milestone for release `1.9.5` in their repository? We don't know what is needed to release this new version without it.

What would be in 1.9.5? See https://github.com/apache/commons-validator/pull/153#issuecomment-1717820757

landsman:
> What would be in 1.9.5? See [#153 (comment)](https://github.com/apache/commons-validator/pull/153#issuecomment-1717820757)

this: https://github.com/apache/commons-beanutils/pull/40 to let me continue in this PR 🙏

garydgregory:
> > What would be in 1.9.5? See [#153 (comment)](https://github.com/apache/commons-validator/pull/153#issuecomment-1717820757)
> 
> this: [apache/commons-beanutils#40](https://github.com/apache/commons-beanutils/pull/40) to let me continue in this PR 🙏

That PR was merged in master for 2.0.0 and it only affects the tests. The runtime still must depend on Collections 3.x because it is in the public API.

PR #216, opened 2024-05-04
Validator CASNumberCheckDigit for Chemical Registry CAS Numbers and ECNumberCheckDigit

homebeaver:
CASNumberCheckDigit - CAS Chemical Abstracts Service RN Registry No

CAS Numbers are unique identification numbers used to identify chemical
substance described in the open scientific literature.

see [wikipedia CAS_Registry_Number](https://en.wikipedia.org/wiki/CAS_Registry_Number)

Also add Validator for EC number - ECN is a unique seven-digit identifier that is assigned to chemical substances

regards

PR #217, opened 2024-05-06
Validator CASNumberCheckDigit for Chemical Registry CAS Numbers and ECNumberCheckDigit

homebeaver:
Hi @garydgregory 
I incidently closed PR #216 - so I open a new one.  Your hints from yesterday are done with last two commits. 

Same description:

CASNumberCheckDigit - CAS Chemical Abstracts Service RN Registry No

CAS Numbers are unique identification numbers used to identify chemical
substance described in the open scientific literature.

see [wikipedia CAS_Registry_Number](https://en.wikipedia.org/wiki/CAS_Registry_Number)

Also add Validator for EC number - ECN is a unique seven-digit identifier that is assigned to chemical substances

regards

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/checkdigit/ModulusCheckDigit.java
+++ b/src/main/java/org/apache/commons/validator/routines/checkdigit/ModulusCheckDigit.java
@@ -187,12 +187,12 @@ public abstract class ModulusCheckDigit extends AbstractCheckDigit implements Se
 
     /**
      * A convenient method to check empty Strings.
-     *  
+     *
      * @param code The code to check
      * @return {@code true} if code is null or empty, otherwise {@code false}
      */
     protected boolean isEmpty(final String code) {
-    	return code == null || code.isEmpty();
+        return code == null || code.isEmpty();
     }
 
     /**
--- a/src/main/java/org/apache/commons/validator/routines/checkdigit/ModulusCheckDigit.java
+++ b/src/main/java/org/apache/commons/validator/routines/checkdigit/ModulusCheckDigit.java
@@ -186,6 +186,16 @@ public abstract class ModulusCheckDigit extends AbstractCheckDigit implements Se
     }
 
     /**
+     * A convenient method to check empty Strings.
+     *  
+     * @param code The code to check
+     * @return {@code true} if code is null or empty, otherwise {@code false}
+     */
+    protected boolean isEmpty(final String code) {
+    	return code == null || code.isEmpty();
+    }
+
+    /**
      * Calculates the <i>weighted</i> value of a character in the
      * code at a specified position.
      * <p>
--- a/src/main/java/org/apache/commons/validator/routines/checkdigit/ModulusCheckDigit.java
+++ b/src/main/java/org/apache/commons/validator/routines/checkdigit/ModulusCheckDigit.java
@@ -186,16 +186,6 @@ public abstract class ModulusCheckDigit extends AbstractCheckDigit implements Se
     }
 
     /**
-     * A convenient method to check empty Strings.
-     *
-     * @param code The code to check
-     * @return {@code true} if code is null or empty, otherwise {@code false}
-     */
-    protected boolean isEmpty(final String code) {
-        return code == null || code.isEmpty();
-    }
-
-    /**
      * Calculates the <i>weighted</i> value of a character in the
      * code at a specified position.
      * <p>
--- a/src/test/java/org/apache/commons/validator/routines/checkdigit/AbstractCheckDigitTest.java
+++ b/src/test/java/org/apache/commons/validator/routines/checkdigit/AbstractCheckDigitTest.java
@@ -35,7 +35,7 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
 /**
- * Luhn Check Digit Test.
+ * Check Digit Test.
  */
 public abstract class AbstractCheckDigitTest {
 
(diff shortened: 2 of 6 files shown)


src/main/java/org/apache/commons/validator/routines/checkdigit/ModulusCheckDigit.java
```diff
@@ -185,6 +185,16 @@ protected int toInt(final char character, final int leftPos, final int rightPos)
...
     }
 
+    /**
+     * A convenient method to check empty Strings.
+     *
+     * @param code The code to check
+     * @return {@code true} if code is null or empty, otherwise {@code false}
+     */
+    protected boolean isEmpty(final String code) {
+        return code == null || code.isEmpty();
```
garydgregory, line 195: Oops, I realized yesterday we already have this: `org.apache.commons.validator.GenericValidator.isBlankOrNull(String)` so replace call sites to this with calls to `isBlankOrNull()`. Sorry about that!

garydgregory (changes requested):
Hello @homebeaver 
Thank you for your update.
I don't see tests that make sure _invalid_ input is in fact invalid. For example, when the check digit is incorrect. 
I think there are also boundary cases that are not covered:
- The longest possible encoding. One max case and cases for each segment
- Same but for the shortest.
TY

homebeaver:
> ...I don't see tests that make sure _invalid_ input is in fact invalid. For example, when the check digit is incorrect.

@garydgregory 

Test `AbstractCheckDigitTest.testIsValidFalse` implemets it and it runs :-)

The theoretical MIN valid EC-Number is 000-001-6; `testIsValidFalse` generates all invalid check digits!  See log:

```
   0 Testing Invalid Check Digit, Code=[0000010]
   1 Testing Invalid Check Digit, Code=[0000011]
   2 Testing Invalid Check Digit, Code=[0000012]
   3 Testing Invalid Check Digit, Code=[0000013]
   4 Testing Invalid Check Digit, Code=[0000014]
   5 Testing Invalid Check Digit, Code=[0000015]
   6 Testing Invalid Check Digit, Code=[0000017]
   7 Testing Invalid Check Digit, Code=[0000018]
   8 Testing Invalid Check Digit, Code=[0000019]
   9 Testing Invalid Check Digit, Code=[000001 ]
   10 Testing Invalid Check Digit, Code=[000001A]
   11 Testing Invalid Check Digit, Code=[000001B]
   12 Testing Invalid Check Digit, Code=[000001C]
   13 Testing Invalid Check Digit, Code=[000001D]
   14 Testing Invalid Check Digit, Code=[000001E]
   15 Testing Invalid Check Digit, Code=[000001F]
   16 Testing Invalid Check Digit, Code=[000001H]
   17 Testing Invalid Check Digit, Code=[000001I]
   18 Testing Invalid Check Digit, Code=[000001J]
   19 Testing Invalid Check Digit, Code=[000001K]
   20 Testing Invalid Check Digit, Code=[000001L]
   21 Testing Invalid Check Digit, Code=[000001M]
   22 Testing Invalid Check Digit, Code=[000001N]
   23 Testing Invalid Check Digit, Code=[000001O]
   24 Testing Invalid Check Digit, Code=[000001P]
   25 Testing Invalid Check Digit, Code=[000001Q]
   26 Testing Invalid Check Digit, Code=[000001R]
   27 Testing Invalid Check Digit, Code=[000001S]
   28 Testing Invalid Check Digit, Code=[000001T]
   29 Testing Invalid Check Digit, Code=[000001U]
   30 Testing Invalid Check Digit, Code=[000001V]
   31 Testing Invalid Check Digit, Code=[000001W]
   32 Testing Invalid Check Digit, Code=[000001X]
   33 Testing Invalid Check Digit, Code=[000001Y]
   34 Testing Invalid Check Digit, Code=[000001Z]
   35 Testing Invalid Check Digit, Code=[000001	]
...
```

src/test/java/org/apache/commons/validator/routines/checkdigit/AbstractCheckDigitTest.java
```diff
@@ -154,7 +154,11 @@ public void testCalculateInvalid() {
                     log.debug("   " + i + " Testing Invalid Check Digit, Code=[" + code + "]");
                 }
                 final String expected = checkDigit(code);
-                final String actual = routine.calculate(removeCheckDigit(code));
+                String codeWithNoCheckDigit = removeCheckDigit(code);
+                if (codeWithNoCheckDigit == null) {
+                    throw new CheckDigitException("Invalid Code=[" + code + "]");
```
garydgregory, line 159: Why not just call `fail()` here?

garydgregory:
> The theoretical MIN valid EC-Number is 000-001-6;

OK, then please add a test for that, something like `testMinimum()`, and the whatever the max value is (if there is one, if not something that looks very large). 

My concern is that if the RE says "this segment can be 1, 2, or 3 digits", then there should be assertions for that.

garydgregory (changes requested):
Hello @homebeaver 
Thank you for your reply. Please see my comments.

homebeaver:
> OK, then please add a test for that, something like `testMinimum()`, and the whatever the max value is (if there is one, if not something that looks very large).

Boundary tests are included in commit deb39a25ac8684334447ebfad4829191ca27c192 [MIN MAX tests]

src/test/java/org/apache/commons/validator/routines/checkdigit/AbstractCheckDigitTest.java
homebeaver: the default invalid test in `class AbstractCheckDigitTest` is
```
protected String[] invalid = { "12345678A" };
```

method `removeCheckDigit` returns null if `REGEX_VALIDATOR.validate(code)` == null. This is the case for "12345678A" because this code is invalid due to its regex.

When using `fail()` as you suggests the test `testCalculateInvalid()` would fail 
but it should `assertTrue(e.getMessage().startsWith("Invalid ") ...` because "12345678A"  IS INVALID!

So we `throw new CheckDigitException("Invalid Code=[" + code + "]");` and catch it in
```
            } catch (final CheckDigitException e) {
                // possible failure messages:
                // Invalid ISBN Length ...
                // Invalid Character[ ...
                // Are there any others?
                assertTrue(e.getMessage().startsWith("Invalid "), "Invalid Character[" + i + "]=" + e.getMessage());
```

NB: `testCalculateInvalid()` is not my idea. I needed a while to understand how it works :-)

Regards

PR #237, opened 2024-06-30
add devcontainer

spotlesscoder:
this leads to a cleaner developer machine:
* easy setup of development environment - just checkout the repo in vscode and run "open in devcontainer"
* dependencies are installed in a docker container and everything can be cleanup up easily after development task is done

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,20 @@
+{
+	"name": "commons-validator",
+	"image": "mcr.microsoft.com/devcontainers/java:1-8-bullseye",
+
+	"features": {
+		"ghcr.io/devcontainers/features/java:1": {
+			"version": "none",
+			"installMaven": "true"
+		}
+	},
+
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"streetsidesoftware.code-spell-checker",
+				"vscjava.vscode-java-pack"
+			]
+		}
+	}
+}


garydgregory:
This is an infinite amount of IDEs, tools, and contraptions out there. I don't think we need to clutter up our repositories with this kind of stuff. YMMV.

spotlesscoder:
The devcontainer spec is nothing specific to VS Code
It would also not interfere with the VS Code installation of anyone until the person starts the devcontainer in VS Code which is always something you have to actively do

spotlesscoder:
I highly doubt that someone uses a devcontainer config without having added it to git.
You would loose all your config if you reinstalled your OS etc.

But even if someone has a local devcontainer config vscode would prompt to simply rebuild the devcontainer

Devcontainers are also not specific to windows. They are just normal docker containers. Any platform that can run docker (e.g. Linux, Windows, Mac - Windows and Mac with docke desktop) can use them

spotlesscoder:
Also Devcontainers are supported by JetBrains and probably also Eclipse
But usually this is opt in. 
If it's not then the IDE sucks

garydgregory:
I've not seen Eclipse generate such a file.

spotlesscoder:
Usually non of the IDEs I know would generate such a file.
You write it once, add it to the source control and use a plugin of your IDE to automatically spin up the container according to the configuration in the devcontainer.json

It seems like in the Eclipse community the devcontainers are not yet that common

PR #271, opened 2024-10-08
VAT Id Number validators for all EU countries

homebeaver:
Hi @garydgregory ,

here is EUgen from Europe. 
With this PR I provide **validation** and check digit **calculation** for VAT Id Numbers [VATIN](https://en.wikipedia.org/wiki/VAT_identification_number) for all countries in the EU. This number is neccessary in the intra comunity trades to apply a zero VAT rate. 
You can find a documentation of the different VATINs structures in my [VATIN wiki (de)](https://github.com/homebeaver/ungueltig/wiki/VATIN)

The validator for all is `VATINValidator` - it implements the [jira issue 494](https://issues.apache.org/jira/projects/VALIDATOR/issues/VALIDATOR-494)

Pls review and merge.
Regards EUGen


homebeaver:
> Is it really necessary to have a separate class for each country? Also DE has a test class but no main class

@sebbASF 

Yes. Each of the 28 countries use a different algoritm to calculate/validate the check digit. 
Germany and Croatia use a standard **ISO/IEC 7064, MOD 11,10**. So the tests are against Modulus11TenCheckDigit
Italy uses LUHN.

src/main/java/org/apache/commons/validator/routines/checkdigit/CheckDigitException.java
```diff
@@ -25,6 +25,16 @@ public class CheckDigitException extends Exception {
...
+    /**
+     * Common message text when code is Null or empty
+     */
+    public static final String MISSING_CODE = "Code is missing";
+
+    /**
+     * Common message text when code sum is zero
+     */
+    public static final String ZREO_SUM = "Invalid code, sum is zero";
+
```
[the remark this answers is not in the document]
homebeaver: Resolved with commit 547c10ee9023600a3547883a1394c8e7abce303f

src/main/java/org/apache/commons/validator/routines/checkdigit/Modulus97CheckDigit.java
```diff
@@ -0,0 +1,177 @@
...
+ * for more details.
+ * </p>
+ *
+ * <p>
+ * This MOD 97-10 module can be used to validate the LEI (Legal Entity Identifier), ICD id 0199
+ * and in VATidBECheckDigit (VATIN_BE), and also in LeitwegValidator
+ *
+ * @since 1.10.0
+ */
+public class Modulus97CheckDigit extends AbstractCheckDigit implements Serializable {
```
garydgregory, line 37: Don't implement Serializable, we are moving away from that IMO, folks that need serialization of some kind can use serialization proxies; see "Effective Java" by J. Block.

src/main/java/org/apache/commons/validator/routines/checkdigit/VATidBECheckDigit.java
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * Constructs a Check Digit routine.
+     */
+    private VATidBECheckDigit() {
+        super();
+    }
+
+    /**
+     * {@inheritDoc}
+     * return expression overridden
```
garydgregory, line 62: Missing `@`

src/main/java/org/apache/commons/validator/routines/checkdigit/VATidBGCheckDigit.java
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * Gets the singleton instance of this validator.
+     * @return A singleton instance of the class.
+     */
+    public static CheckDigit getInstance() {
+        return INSTANCE;
+    }
+
+    /**
+     * there ate three length
```
garydgregory, line 58: Sentences start with a capital letter. If you expect the comment below to be renders line by line, use a `ol` or `ul` HTML tag.

src/main/java/org/apache/commons/validator/routines/checkdigit/VATidCYCheckDigit.java
```diff
@@ -0,0 +1,135 @@
...
+     * <p>For VATID digits are weighted by their position from left to right.</p>
+     *
+     * @param charValue The numeric value of the character.
+     * @param leftPos The position of the character in the code, counting from left to right
+     * @param rightPos The positionof the character in the code, counting from right to left
+     * @return The weighted value of the character.
+     */
+    @Override
+    protected int weightedValue(final int charValue, final int leftPos, final int rightPos) {
+        if (leftPos % 2 == 0) return charValue;
```
garydgregory, line 92: Always use blocks.

src/main/java/org/apache/commons/validator/routines/checkdigit/VATidFRCheckDigit.java
```diff
@@ -0,0 +1,198 @@
...
+
+    static final int LEN = 11; // with Check Digit
+    static final int CHECKDIGIT_LEN = 2;
+
+    private static final String ALPHABET = "0123456789ABCDEFGHJKLMNPQRSTUVWXYZ";
+    private static final int MODULUS_97 = 97;
+    private static final SireneValidator SIRENE_VALIDATOR = SireneValidator.getInstance();
+
+    /**
+     * Constructs a Check Digit routine.
```
garydgregory, line 61: Not helpful IMO. Also, there is no "routine" concept in Java, this constructs a new instance if anything ;-)

garydgregory (changes requested):
Hello @homebeaver 
This is quite a giant PR! I will take time to review and consider. I've made some initial but low-level comments.

src/main/java/org/apache/commons/validator/routines/checkdigit/Modulus97CheckDigit.java
homebeaver: Cannot remove it.
Error:  Failures: 
Error:    Modulus97CheckDigitTest>AbstractCheckDigitTest.testSerialization:300 org.apache.commons.validator.routines.checkdigit.Modulus97CheckDigit error during serialization: java.io.NotSerializableException: org.apache.commons.validator.routines.checkdigit.Modulus97CheckDigit

garydgregory: Hello @homebeaver 
That was a bug in our test ;-) Please rebase on master to pick up https://github.com/apache/commons-validator/commit/fdc8ee9c26f668192b34ec5b955b3adf47f98b2d

src/main/java/org/apache/commons/validator/routines/checkdigit/TidDECheckDigit.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Constructs a modulus Check Digit routine.
+     */
+    private TidDECheckDigit() {
+        super();
+    }
+
+    /**
+     * {@inheritDoc} <br>
```
garydgregory, line 57: Don't leave HTML tags open. You have paragraphs below, this break is likely not needed.

src/main/java/org/apache/commons/validator/routines/checkdigit/VATidLVCheckDigit.java
```diff
@@ -0,0 +1,220 @@
...
+                return cd == Character.getNumericValue(code.charAt(code.length() - 1));
+            }
+            int ccd = calculateNMIN(code.substring(0, LEN - 1), true);
+            return ccd == Character.getNumericValue(code.charAt(code.length() - 1));
+        } catch (final CheckDigitException ex) {
+            return false;
+        }
+    }
+
+    private int vRule1(String code) {
```
garydgregory, line 202: Use `final` where you can, throughout the PR.

src/test/java/org/apache/commons/validator/routines/SireneValidatorTest.java
```diff
@@ -0,0 +1,62 @@
...
+package org.apache.commons.validator.routines;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class SireneValidatorTest {
+
+    private final String[] validFormat = new String[] {
```
garydgregory, line 26: If you define these test fixtures with `Arrays.asList(...)`, you can use `forEach()` to reduce clutter in the tests. YMMV.

src/test/java/org/apache/commons/validator/routines/SireneValidatorTest.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.validator.routines;
+
+import static org.junit.Assert.assertFalse;
```
garydgregory, line 19: Please use JUnit 5 `Assertions`, not JUnit 4 `Assert`.

Please rebase on git master where I finished porting to JUnit 5; there was only one class left to do.

src/test/java/org/apache/commons/validator/routines/VATINValidatorTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+    @Test
+    public void testNull() {
+        assertFalse(VALIDATOR.isValid(null), "isValid(null)");
+    }
+
+    @Test
+    public void testSetDefaultValidator1() {
+        final IllegalStateException thrown = assertThrows(IllegalStateException.class, () -> VALIDATOR.setValidator("GB", 15, "GB", null));
+        assertThat(thrown.getMessage(), is(equalTo("The singleton validator cannot be modified")));
```
garydgregory, line 172: Please, not Hamcrest, not AssertJ, Java is verbose enough as it is without cluttering tests which should be as simple as possible to understand since they also serve to some extent as examples. This is so much simpler to read and maintain:
```
 assertEquals("The singleton validator cannot be modified", thrown.getMessage());
 ```
There is absolutely no need for 2 extra method calls to assert this outcome.

File: src/test/java/org/apache/commons/validator/routines/VATINValidatorTest.java [hunk 12]
```diff
@@ -0,0 +1,251 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests {@link VATINValidator}.
+ */
+public class VATINValidatorTest {
+
+    private static final VATINValidator VALIDATOR = VATINValidator.getInstance();
+
+    // Eclipse 3.6 allows you to turn off formatting by placing a special comment, like
```
garydgregory, line 40: You don't need a comment about Eclipse. We use this `formatter` comment all over Commons 😉

src/test/java/org/apache/commons/validator/routines/VATINValidatorTest.java
```diff
@@ -0,0 +1,251 @@
...
+        assertNull(validator.setValidator("DE", -1, "", null), "no longer present");
+    }
+
+    private static final String INVALID_LENGTH = "Invalid length parameter, must be in range 10 to 16 inclusive:";
+
+    @Test
+    public void testSetValidatorLen35() {
+        final VATINValidator validator = new VATINValidator();
+        final IllegalArgumentException thrown = assertThrows(IllegalArgumentException.class, () -> validator.setValidator("DE", 35, "DE", null));
+//        System.out.println("testSetValidatorLen35 : thrown.getMessage():" + thrown.getMessage());
```
garydgregory, line 201: Remove comment.

src/test/java/org/apache/commons/validator/routines/checkdigit/VATidFRCheckDigitTest.java
```diff
@@ -0,0 +1,132 @@
...
+    protected void setUp() {
+        checkDigitLth = VATidFRCheckDigit.CHECKDIGIT_LEN;
+        routine = VATidFRCheckDigit.getInstance();
+
+        valid = new String[] {"00300076965"
+            , "55502090897"
+            , "40303265045"
+            , "23334175221"
+            , "06399859412"
+//            , "K7399859412" // SLRL ALTEA EXPERTISE COMPTABLE => test in VATINValidatorTest
```
garydgregory, line 47: Is this work in progress?

src/test/java/org/apache/commons/validator/routines/checkdigit/VATidFRCheckDigitTest.java
```diff
@@ -0,0 +1,132 @@
...
+            , "83404833048"
+            , "11123456782" // konstruiert: L'identifiant n'existe pas dans le répertoire Sirene.
```
garydgregory, line 49: This comment needs to be more helpful to maintainers, as it the one above.

src/test/java/org/apache/commons/validator/routines/checkdigit/VATidFRCheckDigitTest.java
```diff
@@ -0,0 +1,132 @@
...
+        // test invalid check digit values
+        final String[] invalidCheckDigits = createInvalidCodes(valid);
+        final Map<String, List<String>> icdmap = new Hashtable<String, List<String>>();
+        for (int i = 0; i < invalidCheckDigits.length; i++) {
+            if (log.isDebugEnabled()) {
+                log.debug("   " + i + " Testing Invalid Check Digit, Code=[" + invalidCheckDigits[i] + "]");
+            }
+            boolean res = routine.isValid(invalidCheckDigits[i]);
+            if (res) {
+//                log.warn("   " + i + " Testing Invalid Check Digit, Code=[" + invalidCheckDigits[i] + "] is true, expected false.");
```
garydgregory, line 106: Remove // commented out code, these look like work-in-progress.

src/test/java/org/apache/commons/validator/routines/checkdigit/VATidLVCheckDigitTest.java
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ */
+package org.apache.commons.validator.routines.checkdigit;
+
+import org.junit.jupiter.api.BeforeEach;
+
+/**
+ * LV VAT Id Check Digit Tests.
+ */
+/*
```
garydgregory, line 24: Merge comments.

src/test/java/org/apache/commons/validator/routines/checkdigit/VATidROCheckDigitTest.java
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+package org.apache.commons.validator.routines.checkdigit;
+
+import org.junit.jupiter.api.BeforeEach;
+
+/**
+ * RO VAT Id Check Digit Tests.
+ */
+/*
```
garydgregory, line 24: Merge comments.

src/test/java/org/apache/commons/validator/routines/checkdigit/VATidSKCheckDigitTest.java
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+package org.apache.commons.validator.routines.checkdigit;
+
+import org.junit.jupiter.api.BeforeEach;
+
+/**
+ * SK VAT Id Check Digit Tests.
+ */
+/*
```
garydgregory, line 24: Merge comments.

src/test/java/org/apache/commons/validator/routines/checkdigit/VATidXICheckDigitTest.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/**
+ * XI VAT Id Check Digit Tests.
+ */
+/*
```
garydgregory, line 33: Merge comments.

src/test/java/org/apache/commons/validator/routines/checkdigit/VATidXICheckDigitTest.java
```diff
@@ -0,0 +1,173 @@
...
+              , "613451470"
+              , "107328000"
+              , "766800804"
+              , "980780684"
+              , "888801276"
+              , "888850259"
+              , "888851256"
+              , "816137833"
+              , "562235945"
+              /* new style 9755 cannot be tested here
```
garydgregory, line 79: What does "new style" mean? Think about maintenance in these comments ;-)
Where can it be tested then?

garydgregory (changes requested):
Hello @homebeaver 

Thank you for you updates.

In general, where does the test data come from? If there are resources on the web, it would be nice to provide URLs in the Javadoc for provenance.

I've added comments scattered throughout.

Please rebase on git master.

TY!

garydgregory:
Hi @homebeaver 

A nice to have item would be an update to the `package-info.java` file for the package in play.

homebeaver:
Hi @garydgregory 
where does the test data come from?

Different - I documented it in [wiki/VATIN](https://github.com/homebeaver/ungueltig/wiki/VATIN) - in German!
You'll find examples in the Details-link:

grafik

src/main/java/org/apache/commons/validator/routines/checkdigit/Modulus97CheckDigit.java
homebeaver: after rebase and rempve implement Serializable:
```
[INFO] --- spotbugs-maven-plugin:4.8.6.4:check (default-cli) @ commons-validator ---
[INFO] BugInstance size is 1
[INFO] Error size is 0
[INFO] Total bugs: 1
Error:  Medium: Class org.apache.commons.validator.routines.CodeValidator defines non-transient non-serializable instance field checkdigit [org.apache.commons.validator.routines.CodeValidator] In CodeValidator.java SE_BAD_FIELD
[
```

garydgregory: Ah, I see: Because `CodeValidator implements Serializable` and `CodeValidator` has a `CheckDigit` field, we are stuck with `CheckDigit` implementing `Serializable` if we want to honor the `Serializable` contract on `CodeValidator`. I guess we'll have to drop serialization in a major version bump if we want to go that way.

TY for cycling through this one @homebeaver ! 😄

homebeaver: I revert the commit 8add96b5939ef1cc2d3ab3ff8d774ee48f194d73

src/test/java/org/apache/commons/validator/routines/checkdigit/VATidXICheckDigitTest.java
homebeaver: I`ve explained "new style" in the file. Here in short:
- each VAT code in GB and Nothern Ireland (XI) has two check digits - both are valid
- one calculated with MOD97 algorithm - this is called "old style" (this is the result of calculate method)
- the second calculated  with MOD9755 - this is called "new style"

homebeaver:
> A nice to have item would be an update to the `package-info.java` file for the package

Hi @garydgregory 

updated the `package-info.java`  in `org.apache.commons.validator.routines
`
Regards - next week I'm on holiday :-)
EUGen

raboof:
I'm afraid the NL validation is incomplete: my VAT ID, NL004350351B91, is not accepted while it is in fact valid (confirmed with the online VIES check). I suspect this may have to do with a change in 2020 or so: before that time, the btw-id for self-employed people consisted of 'NL'+their BSN (the Dutch SSN) + "B" + two digits. While 'leaking' a SSN isn't as problematic in NL as it is in some other countries, it's not great, so they stopped doing that. Apparently the BSN adheres to the `MOD-11#i` pattern, while the new number does not. I can share a second number that fails the validator but is correct according to VIES privately on request.

raboof:
> Is the new NL VAT validation scheme officially documented anywhere?

I looked around a bit but didn't find anything - the sources I found just say 'NL, 9 digits, B, 2 digits', no further structure described.

> It seems to me that even with access to official documentation, it is going to be a mammoth job keeping track of changes.
>
> Tracking updates to IBAN and Domain entries is time-consuming enough, and they each have a single authoritative source.
> 
> Sorry, but I think the VAT validation proposal is not a suitable candidate for Commons Validator.

OTOH, you could also say that this is a good motivation to share this effort. Perhaps we should make sure that there are at least 3 or so parties that are planning to use this in production (and share notes when rules change) before including it?

garydgregory:
I support @raboof's idea 👍

homebeaver:
> ... VAT ID, NL004350351B91, is not accepted while it is in fact valid (confirmed with the online VIES check). I suspect this may have to do with a change in 2020 or so: before that time, the btw-id for self-employed people consisted of 'NL'+their BSN (the Dutch SSN) + "B" + two digits. ...

@raboof thank you for testing. Indeed, my algorithm use MOD 11 with $${ G_i = i }$$ - this was changed in 2020: 
There are two groups with VAT-numbers:
• Modulo 11 VAT-ID numbers;
• Modulo 97 VAT-ID numbers.

I´ll correct this.

regards EUGen H.

homebeaver:
> I looked around a bit but didn't find anything ...

@raboof  - I found this [BMF_UID_Konstruktionsregeln-Nov-2020.pdf bmf.gv.at](https://www.bmf.gv.at/dam/bmfgvat/finanzonline/fuer-softwarehersteller/sonstige-erklaerungen/Zusammenfassende-Meldung/BMF_UID_Konstruktionsregeln_Stand_November-2020.pdf)

homebeaver:
> ... it does not check that the ID is actually in use. So I wonder what the use case is?

Compate VATID to IBAN: validating an IBAN does not check if the IBAN is actually in use.
Use VIES for checking if a VATIN is actually in use. It delegates the check to the national systems, but is not avaiable 7/24. We use it for valid VATINS only when creating an invioce.

The use cases: 
european standard EN16931-1:2017 for electronic invoices needs an valid VATIN of the cusomer. This number is neccessary in the intra comunity trades to apply a zero VAT rate.

In my case I migrate ERP data from system A to system B. There are a lot of unchecked VATINs and VIES is not the tool of choice for checking a mass. When a shop or company discontinues VIES returns invalid (not invalid since...), but invoices are stored a while for auditing.

raboof:
> > I looked around a bit but didn't find anything ...
> 
> @raboof - I found this [BMF_UID_Konstruktionsregeln-Nov-2020.pdf bmf.gv.at](https://www.bmf.gv.at/dam/bmfgvat/finanzonline/fuer-softwarehersteller/sonstige-erklaerungen/Zusammenfassende-Meldung/BMF_UID_Konstruktionsregeln_Stand_November-2020.pdf)

Nice find! I checked and can confirm that both my own btw-id and the one I won't disclose are accepted by the updated implementation.

To justify the additional maintenance burden it'd still be good to find one or two other parties that can confirm they'd use this component in production and help keep it up-to-date before merging it into the main library.

sebbASF (changes requested):
These look like bugs

src/main/java/org/apache/commons/validator/routines/checkdigit/VATidBECheckDigit.java
```diff
@@ -0,0 +1,104 @@
...
+     * {@inheritDoc}
+     * <p>
+     * Overridden because the check digits are calculated as 97 - modulusResult
+     * </p>
+     */
+    @Override
+    public String calculate(final String code) throws CheckDigitException {
+        if (GenericValidator.isBlankOrNull(code)) {
+            throw new CheckDigitException(CheckDigitException.MISSING_CODE);
+        }
```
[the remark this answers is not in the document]
homebeaver: You can find the explanation in `package-info` of  package `org.apache.commons.validator.routines.checkdigit`:


_This package contains Check Digit validation/calculation routines._

_Note that these **do not validate the input for length** or syntax. Such validation is performed by the org.apache.commons.validator.routines.XYZValidator classes._

`VATINValidator` class validates against reg-ex `"[0-1]\\d{9}"`

In checkdigit package I use LEN only if it is absolute necessary for the implementation of the check.

src/main/java/org/apache/commons/validator/routines/checkdigit/VATidLUCheckDigit.java
```diff
@@ -0,0 +1,110 @@
...
+import org.apache.commons.validator.GenericTypeValidator;
+import org.apache.commons.validator.GenericValidator;
+
+/**
+ * Luxemburg VAT identification number (VATIN) Check Digit calculation/validation.
+ * <p>
+ * Numéro d'identification à la taxe sur la valeur ajoutée {@code 123456pp}.
+ * </p>
+ * <p>
+ * The check digits are calculated as 89 - MOD 89
```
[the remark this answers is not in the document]
homebeaver: Resolved with Commit 5e1c022

src/main/java/org/apache/commons/validator/routines/checkdigit/VATidBECheckDigit.java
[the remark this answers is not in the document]
homebeaver: To be consistent with `VATidBECheckDigit`

I change in `VATidLUCheckDigit` from `MIN_LEN` which is 4 to the stronger `CHECKDIGIT_LEN` which is 2.

Done in Commit dc6e9c1

homebeaver:
> These look like bugs

No. Imprecise comments + inconsistent code. Both resolved. Tkx

ecki (approved):

src/main/java/org/apache/commons/validator/routines/checkdigit/VATidBECheckDigit.java
```diff
@@ -0,0 +1,104 @@
...
+    public boolean isValid(final String code) {
+        if (GenericValidator.isBlankOrNull(code)) {
+            return false;
+        }
+        if (code.length() <= CHECKDIGIT_LEN) {
+            return false;
+        }
+        final String check = code.substring(code.length() - CHECKDIGIT_LEN);
+        try {
+            final String cd = calculate(code.substring(0, code.length() - CHECKDIGIT_LEN)); // throws CheckDigitException
```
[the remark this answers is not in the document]
homebeaver: I plan a new PR with [implementation for all modules defined in ISO/IEC 7064](https://issues.apache.org/jira/browse/VALIDATOR-481) when this PR is finished/merged. 
This would change some of the base classes, anyway.

src/main/java/org/apache/commons/validator/routines/checkdigit/VATidBECheckDigit.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * {@inheritDoc}
+     * <p>
+     * Overridden because the check digits are calculated as 97 - modulusResult
+     * </p>
+     */
+    @Override
+    public String calculate(final String code) throws CheckDigitException {
+        if (GenericValidator.isBlankOrNull(code)) {
+            throw new CheckDigitException(CheckDigitException.MISSING_CODE);
```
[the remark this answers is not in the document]
homebeaver: I plan a new PR with [implementation for all modules defined in ISO/IEC 7064](https://issues.apache.org/jira/browse/VALIDATOR-481) when this PR is finished/merged. 
This would change some of the base classes, anyway.

garydgregory:
I wouldn't except any logging from this component.

homebeaver:
I've removed logging.

There is one class with logging left : `TidDECheckDigit`. The reason is the specification - the link to it is in the class header. The spec is in german. In some cases valid tids should not be used - and I warn with the message recomended in the spec. See page 6+7 of the spec or screenshot:
grafik

> I wouldn't except any logging from this component.

@garydgregory 
To satisfy you I can remove the class from this PR

sebbASF (changes requested):
As per comments

homebeaver:
Hi @sebbASF , @sebbASF 

when merging apache/master to my repo branch I got an error in `pom.xml`

```
- cvc-elt.1.a: Cannot find the declaration of element 'project'.
- Downloading external resources is disabled.
```

The solution is **https** instead http in pom element 'project'

I resolve this in my pull request and hope you accept this pull. It is waiting to be accepted for more then one year!

regards EUGen

sebbASF (changes requested):

homebeaver:
tkx @raboof.
Changes can be cleanly merged. @garydgregory

garydgregory:
> > I resolve this in my pull request and hope you accept this pull. It is waiting to be accepted for more then one year!
> 
> As mentioned earlier in [#271 (comment)](https://github.com/apache/commons-validator/pull/271#issuecomment-2623175681): keeping this up-to-date looks like quite a maintenance burden. This means the work is valuable to share - but only if we're confident that there will be someone looking after this code after it's merged. I assume you're using this and would be ready to take some of that maintainership on yourself, but it'd be good to have one or two other people showing up.

There has been no reply from @homebeaver to your comment @raboof, which doesn't inspire confidence that this will be maintained.

garydgregory (commented):
@homebeaver At least one of my comments hasn't been addressed; it was marked resolved, which is not appropriate IMO, so I stopped looking. Beyond that, I'm not sure there is consensus to bring this in due to the maintenance burden. What do others think?

sebbASF:
I agree that this is likely to become too much of a maintenance burden.

Seems to me this would be better as a separate project (with its own package name).
This could of course use existing Validator methods where appropriate.

It might be worth considering adding some or all of the new ModulusCheckDigit classes, if the algorithm is standardised and in general use.

PR #293, opened 2024-12-17
Improve IBANValidatorTest - use registry from swift

slawekjaranowski:
- add iban_registry_v98.txt from swift site
- use parametrized test


src/test/java/org/apache/commons/validator/routines/IBANValidatorTest.java
```diff
@@ -309,95 +248,198 @@ private static String formatToRE(final String type, final int len) {
...
+    @ParameterizedTest
+    @MethodSource("ibanRegistryV98SourceExamples")
+    void exampleAccountsShouldBeValid(String countryName, String example) {
+        Assumptions.assumeFalse(INVALID_IBAN_FIXTURES.contains(example), "Skip invalid example: " + example + " for " + countryName);
+        assertTrue(IBANValidator.getInstance().isValid(example), "IBAN validator returned false for " + example + " for " + countryName);
     }
 
     @Test
-    public void testGetRegexValidatorPatterns() {
+    void testGetRegexValidatorPatterns() {
```
garydgregory, line 373: Hello @slawekjaranowski 
Please don't make cosmetic changes, it makes it harder and takes longer to review a PR (FWIW, I don't agree with the JUnit style guideline on this kind of change). Ideally, a PR should only contain required changes. TY!

slawekjaranowski: ok, I will revert such

slawekjaranowski: @garydgregory  - done

garydgregory:
TY @slawekjaranowski 😊 
How soon do you need a release?

slawekjaranowski:
Hi @garydgregory thanks for asking ... 

I would like do https://issues.apache.org/jira/browse/VALIDATOR-497 and https://issues.apache.org/jira/browse/VALIDATOR-496
after it I think release will be ok 😄

garydgregory:
The build is broken btw.

slawekjaranowski:
> The build is broken btw.

but only on JDK 23+ I hope that this PR doesn't break more ...

garydgregory:
The failure was due to a checkstyle error. The other failures for Java 23+ are for builds marked experimental, those are known issues.

PR #294, opened 2024-12-18
[VALIDATOR-496] IBANValidator: Fix IBAN 'c' pattern according to specification

slawekjaranowski:


slawekjaranowski:
@sebbASF thanks ... so I close it 😄

PR #302, opened 2024-12-26
IBANValidator: remove unneeded expression

slawekjaranowski:


src/main/java/org/apache/commons/validator/routines/IBANValidator.java
```diff
@@ -101,10 +101,6 @@ public Validator(final String countryCode, final int ibanLength, final String re
             if (ibanLength > MAX_LEN || ibanLength < MIN_LEN) {
                 throw new IllegalArgumentException("Invalid length parameter, must be in range " + MIN_LEN + " to " + MAX_LEN + " inclusive: " + ibanLength);
             }
-            final String regex = countryCode + regexWithoutCC;
```
slawekjaranowski, line 104: `regex` variable is not used in other place

src/main/java/org/apache/commons/validator/routines/IBANValidator.java
```diff
@@ -101,10 +101,6 @@ public Validator(final String countryCode, final int ibanLength, final String re
...
-            if (!regex.startsWith(countryCode)) {
```
slawekjaranowski, line 105: This is always false

garydgregory:
TY @slawekjaranowski 👍

PR #307, opened 2024-12-28
Optimize the cleanup of the map.

lctking:
**Optimize the cleanup of the map.
Check if the map is empty. If it's not empty, call the native cleanup method of the map; if it is empty, assign a value.**

- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible but is a best-practice.
- [x] Each commit in the pull request should have a meaningful subject line and body. Note that commits might be squashed by a maintainer on merge.

### DIFF
--- a/src/main/java/org/apache/commons/validator/Validator.java
+++ b/src/main/java/org/apache/commons/validator/Validator.java
@@ -196,7 +196,9 @@ public class Validator implements Serializable {
     public void clear() {
         formName = null;
         fieldName = null;
-        parameters = new HashMap<>();
+        if(parameters != null){
+            parameters.clear();
+        }else parameters = new HashMap<>();
         page = 0;
     }
 
--- a/src/main/java/org/apache/commons/validator/Validator.java
+++ b/src/main/java/org/apache/commons/validator/Validator.java
@@ -90,29 +90,29 @@ public class Validator implements Serializable {
     /**
      * The Validator Resources.
      */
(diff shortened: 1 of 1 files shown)


garydgregory:
@lctking
-1 You did not follow the instructions, even though you checked the box for "Run a successful build using the default [Maven](https://maven.apache.org/) goal with mvn; that's mvn on the command line by itself."

src/main/java/org/apache/commons/validator/Validator.java
```diff
@@ -196,7 +196,11 @@ public Validator(final ValidatorResources resources, final String formName, fina
     public void clear() {
         formName = null;
         fieldName = null;
-        parameters = new HashMap<>();
+        if(parameters == null){
```
garydgregory, line 199: How can `parameters` ever be null?

garydgregory:
@lctking 
Thank you for the PR but as noted, it cannot be accepted as is. Instead, I pushed new getters to git master and deprecated access to the instance variables directly. You are credited in changes.xml.

src/main/java/org/apache/commons/validator/Validator.java
pzygielo: - https://github.com/apache/commons-validator/commit/2d983f2f2a5e5523754ff88c611250204e2dbf66#r150814815

PR #308, opened 2024-12-29
Fix grammar issues in the JavaDoc comments

sigee:
- Comparison requires 'than', not 'then'.
- Use 'a' instead of 'an' or vice versa.
- The abbreviations 'i.e.' and 'e.g.' requires two periods.
- Some words are spelled with hyphen
- Some words spelled as one word instead of two
- Repeated words (e.g. 'the the')
- Missing comma
- Plural instead of possessive
- Possessive instead of "it's" (short for "it is")

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/ISBNValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/ISBNValidator.java
@@ -28,7 +28,7 @@ import org.apache.commons.validator.routines.checkdigit.ISBN10CheckDigit;
  * <p>
  * This validator validates the code is either a valid ISBN-10
  * (using a {@link CodeValidator} with the {@link ISBN10CheckDigit})
- * or a valid ISBN-13 code (using a {@link CodeValidator} with the
+ * or a valid ISBN-13 code (using a {@link CodeValidator} with
  * the {@link EAN13CheckDigit} routine).
  * <p>
  * The {@code validate()} methods return the ISBN code with formatting
@@ -61,7 +61,7 @@ import org.apache.commons.validator.routines.checkdigit.ISBN10CheckDigit;
  *     <li>979-10, 979-11, 979-12 are assigned to the ISBN agency</li>
  * </ul>
  * All other 979 prefixed EAN-13 numbers have not yet been assigned to an agency. The
- * validator validates all 13 digit codes with 978 or 979 prefixes.
+ * validator validates all 13-digit codes with 978 or 979 prefixes.
  *
  * @since 1.4
  */
--- a/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
@@ -325,7 +325,7 @@ public class UrlValidator implements Serializable {
 
     /**
      * Tests whether the given flag is off.  If the flag is not a power of 2
-     * (ie. 3) this tests whether the combination of flags is off.
+     * (i.e. 3) this tests whether the combination of flags is off.
      *
      * @param flag Flag value to check.
      * @return whether the specified flag value is off.
@@ -336,7 +336,7 @@ public class UrlValidator implements Serializable {
 
     /**
      * Tests whether the given flag is on.  If the flag is not a power of 2
-     * (ie. 3) this tests whether the combination of flags is on.
+     * (i.e. 3) this tests whether the combination of flags is on.
      *
      * @param flag Flag value to check.
      * @return whether the specified flag value is on.
@@ -388,7 +388,7 @@ public class UrlValidator implements Serializable {
      * Returns true if the authority is properly formatted.  An authority is the combination
      * of hostname and port.  A {@code null} authority value is considered invalid.
      * Note: this implementation validates the domain unless a RegexValidator was provided.
-     * If a RegexValidator was supplied and it matches, then the authority is regarded
+     * If a RegexValidator was supplied, and it matches, then the authority is regarded
      * as valid with no further checks, otherwise the method checks against the
      * AUTHORITY_PATTERN and the DomainValidator (ALLOW_LOCAL_URLS)
      * @param authority Authority value to validate, allows IDN
@@ -495,7 +495,7 @@ public class UrlValidator implements Serializable {
     }
 
     /**
-     * Returns true if the query is null or it's a properly formatted query string.
+     * Returns true if the query is null, or it's a properly formatted query string.
      * @param query Query value to validate.
      * @return true if query is valid.
      */
@@ -507,9 +507,9 @@ public class UrlValidator implements Serializable {
     }
 
     /**
-     * Validate scheme. If schemes[] was initialized to a non null,
+     * Validate scheme. If schemes[] was initialized to a non-null,
      * then only those schemes are allowed.
-     * Otherwise the default schemes are "http", "https", "ftp".
+     * Otherwise, the default schemes are "http", "https", "ftp".
      * Matching is case-blind.
      * @param scheme The scheme to validate.  A {@code null} value is considered
      * invalid.
--- a/src/main/java/org/apache/commons/validator/routines/package-info.java
+++ b/src/main/java/org/apache/commons/validator/routines/package-info.java
@@ -420,8 +420,8 @@
  * of the matched <em>groups</em> or {@code null} if invalid.</li>
  * </ul>
  * </li>
- * <li><strong>Case Sensitivity</strong> - matching can be done in either a <i>case
- * sensitive</i> or <em>case in-sensitive</em> way.</li>
+ * <li><strong>Case Sensitivity</strong> - matching can be done in either a
+ * <i>case-sensitive</i> or <em>case in-sensitive</em> way.</li>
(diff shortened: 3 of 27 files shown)


raboof (approved):
while I don't find all of these changes improvements (in particular I don't think those comma's are necessary/helpful), some are clear fixes, so this seems like a net improvement +1.

garydgregory:
I would replace "i.e." and "e.g." and other Latin abbreviation with their English equivalent. Back when I was writing books, our style guide did allow their use. Most non-native speakers (and some native) don't know the difference to boot.

sigee:
@garydgregory I am not a native English speaker, and I am unsure what you mean by replacing them with their English equivalents. I understand the meanings of 'i.e.' and 'e.g.' in English, but I am not certain whether you are referring to their full expressions, such as 'that is' for 'i.e.' or 'for example' for 'e.g.' Wouldn't that be somewhat excessive?

garydgregory:
Hello @sigee and happy new year
Yes, that's exactly what I mean.
I plan on doing such replacement whenever I see them.

sigee:
I searched for them ("e.g.", "i.e.") in the main java files and replaced them. (I hope, not messed up too much)

garydgregory:
Oh please don't make me review 32 files. The original PR was a single file? Or two? When I do this, I usually read the Javadoc to make sure it actually sounds good, as opposed to search, replace, and hope for the best. I have no idea what you did of course.

sigee:
I am sorry for the extra work, but the original change contained 27 files already. So just 5 more files changed.

sigee:
@sebbASF Thank you for your review. I did the changes you requested. Only the last one is left, where the original text was "i.e." and I changed it to "that is", but you said "Does not make sense to me as either "that is" or "for example.".

For me, "taht is" seems ok, because there are all of the relevant validators listed as explanation.

src/main/java/org/apache/commons/validator/util/Flags.java
```diff
@@ -176,7 +176,7 @@ public void turnOffAll() {
     }
 
     /**
-     * Turns on the given flag.  If the flag is not a power of 2 (ie. 3) this
+     * Turns on the given flag. If the flag is not a power of 2 (that is, 3) this
```
raboof, line 179:
```suggestion
     * Turns on the given flag. If the flag is not a power of 2 (for example, 3) this
```

src/main/java/org/apache/commons/validator/util/Flags.java
```diff
@@ -159,7 +159,7 @@ public String toString() {
     }
 
     /**
-     * Turns off the given flag.  If the flag is not a power of 2 (ie. 3) this
+     * Turns off the given flag. If the flag is not a power of 2 (that is, 3) this
```
raboof, line 162:
```suggestion
     * Turns off the given flag. If the flag is not a power of 2 (for example, 3) this
```

src/main/java/org/apache/commons/validator/util/Flags.java
```diff
@@ -132,8 +132,8 @@ public boolean isOff(final long flag) {
     }
 
     /**
-     * Tests whether the given flag is on.  If the flag is not a power of 2
-     * (ie. 3) this tests whether the combination of flags is on.
+     * Tests whether the given flag is on. If the flag is not a power of 2
+     * (that is, 3) this tests whether the combination of flags is on.
```
raboof, line 136:
```suggestion
     * (for example, 3) this tests whether the combination of flags is on.
```

src/main/java/org/apache/commons/validator/util/Flags.java
```diff
@@ -121,8 +121,8 @@ public int hashCode() {
     }
 
     /**
-     * Tests whether the given flag is off.  If the flag is not a power of 2
-     * (ie. 3) this tests whether the combination of flags is off.
+     * Tests whether the given flag is off. If the flag is not a power of 2
+     * (that is, 3) this tests whether the combination of flags is off.
```
raboof, line 125:
```suggestion
     * (for example, 3) this tests whether the combination of flags is off.
```

src/main/java/org/apache/commons/validator/routines/package-info.java
```diff
@@ -420,8 +420,8 @@
  * of the matched <em>groups</em> or {@code null} if invalid.</li>
  * </ul>
  * </li>
- * <li><strong>Case Sensitivity</strong> - matching can be done in either a <i>case
- * sensitive</i> or <em>case in-sensitive</em> way.</li>
+ * <li><strong>Case Sensitivity</strong> - matching can be done in either a
+ * <i>case-sensitive</i> or <em>case in-sensitive</em> way.</li>
```
raboof, line 424:
```suggestion
 * <i>case-sensitive</i> or <em>case-insensitive</em> way.</li>
```

src/main/java/org/apache/commons/validator/routines/package-info.java
```diff
@@ -369,7 +369,7 @@
  * </pre>
  * <p>
  * <strong>Note</strong>: in theory the other numeric validators besides
- * <em>Float</em>, <em>Double</em> and <em>BigDecimal</em> (i.e. <em>Byte</em>,
+ * <em>Float</em>, <em>Double</em> and <em>BigDecimal</em> (that is, <em>Byte</em>,
```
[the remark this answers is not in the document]
raboof: the question is whether "<em>Byte</em>, <em>Short</em>, <em>Integer</em>, <em>Long</em> and <em>BigInteger</em>" is the full list of "the other numeric validators besides <em>Float</em>, <em>Double</em> and <em>BigDecimal</em>", or just a list of examples.

Looking at the code this is indeed the full list, so `that is` seems correct to me.

src/main/java/org/apache/commons/validator/routines/checkdigit/VerhoeffCheckDigit.java
```diff
@@ -96,7 +96,7 @@ public String calculate(final String code) throws CheckDigitException {
      * @param code The code to calculate the checksum for.
      * @param includesCheckDigit Whether the code includes the Check Digit or not.
      * @return The checksum value
-     * @throws CheckDigitException if the code contains an invalid character (i.e. not numeric)
+     * @throws CheckDigitException if the code contains an invalid character (that is not numeric)
```
raboof, line 99:
```suggestion
     * @throws CheckDigitException if the code contains an invalid character (that is, a non-numeric character)
```

src/main/java/org/apache/commons/validator/routines/checkdigit/ISBNCheckDigit.java
```diff
@@ -68,7 +68,7 @@ public ISBNCheckDigit() {
      * @return The ISBN-10 check digit if the length is 9 or an ISBN-13
      * check digit if the length is 12.
      * @throws CheckDigitException if the code is missing, or an invalid
-     * length (i.e. not 9 or 12) or if there is an error calculating the
+     * length (that is not 9 or 12) or if there is an error calculating the
```
raboof, line 71:
```suggestion
     * length (that is, not 9 or 12) or if there is an error calculating the
```

src/main/java/org/apache/commons/validator/routines/UrlValidator.java
```diff
@@ -131,7 +131,7 @@ public class UrlValidator implements Serializable {
 
     private static final String AUTHORITY_REGEX =
             "(?:\\[(" + IPV6_REGEX + ")\\]|(?:(?:" + USERINFO_FIELD_REGEX + ")?([" + AUTHORITY_CHARS_REGEX + "]*)))(?::(\\d*))?(.*)?";
-    //             1                                 e.g. user:pass@           2                                       3       4
+    //             1                                 for example, user:pass@           2                                       3       4
```
raboof, line 134: I think you need to remove some spaces here to make the numbers line up with the regex again

src/main/java/org/apache/commons/validator/routines/ISBNValidator.java
```diff
@@ -220,7 +220,7 @@ public boolean isValidISBN13(final String code) {
      * Check the code is either a valid ISBN-10 or ISBN-13 code.
      * <p>
      * If valid, this method returns the ISBN code with
-     * formatting characters removed (i.e. space or hyphen).
+     * formatting characters removed (such as space or hyphen).
```
raboof, line 223: are you sure other formatting characters than spaces or hyphens are indeed removed?

sigee: Well, formatting characters are called SEP inside the regex pattern.
Based on the SEP constant it could be a hyphen or **any whitespace character**, not only space.
`    private static final String SEP = "(?:\\-|\\s)";`
As I know:
- Space ( )
- Tab (\t)
- Newline (\n)
- Carriage return (\r)
- Form feed (\f)

raboof: Thanks for confirming, that means replacing `i.e.` with `such as` is a documentation bugfix in this case :+1:

raboof (approved):

PR #312, opened 2025-01-15
[VALIDATOR-498] IBAN account checksum proper validation for CZ

rkosmowski:

- [ ] Read the [contribution guidelines](CONTRIBUTING.md) for this project.
- [ ] Run a successful build using the default [Maven](https://maven.apache.org/) goal with `mvn`; that's `mvn` on the command line by itself.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible but is a best-practice.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Each commit in the pull request should have a meaningful subject line and body. Note that commits might be squashed by a maintainer on merge.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/IBANValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/IBANValidator.java
@@ -22,13 +22,15 @@ import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
+import org.apache.commons.validator.routines.checkdigit.CheckDigit;
+import org.apache.commons.validator.routines.checkdigit.CNBCheckDigit;
 import org.apache.commons.validator.routines.checkdigit.IBANCheckDigit;
 
 /**
  * IBAN Validator.
  * <p>
  * The validator includes a default set of formats derived from the IBAN registry at
- * https://www.swift.com/standards/data-standards/iban.
+ * <a href="https://www.swift.com/standards/data-standards/iban">https://www.swift.com/standards/data-standards/iban</a>.
  * </p>
  * <p>
  * This can get out of date, but the set can be adjusted by creating a validator and using the
@@ -58,7 +60,6 @@ public class IBANValidator {
         /**
          * The minimum length does not appear to be defined by the standard.
          * Norway is currently the shortest at 15.
-         *
          * There is no standard for BBANs; they vary between countries.
          * But a BBAN must consist of a branch id and account number.
          * Each of these must be at least 2 chars (generally more) so an absolute minimum is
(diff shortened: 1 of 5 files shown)


src/test/java/org/apache/commons/validator/routines/IBANValidatorTest.java
```diff
@@ -55,7 +55,7 @@
 /**
  * Tests {@link IBANValidator}.
  */
-public class IBANValidatorTest {
+class IBANValidatorTest {
```
slawekjaranowski, line 58: Not connected with change

src/test/java/org/apache/commons/validator/routines/checkdigit/CNBCheckDigitTest.java
slawekjaranowski, line 1: Missing license header

src/test/java/org/apache/commons/validator/routines/IBANValidatorTest.java
garydgregory: I agree. Please keep your changes to a minimum.

garydgregory:
@rkosmowski 
Note that the checklist asks you to run "mvn" by itself which runs all checks.

garydgregory:
Jira [VALIDATOR-498](https://issues.apache.org/jira/projects/VALIDATOR/issues/VALIDATOR-498)

src/main/java/org/apache/commons/validator/routines/IBANValidator.java
```diff
@@ -112,6 +118,11 @@ public Validator(final String countryCode, final int ibanLength, final String re
             this.regexValidator = new RegexValidator(regexList);
         }
 
+        public Validator withSpecificCheckDigit(CheckDigit specificAccountCheckDigit) {
```
slawekjaranowski, line 121: public method should have a documentation ... 
please add javadoc for it

slawekjaranowski (approved):
LGTM

slawekjaranowski:
@sebbASF @garydgregory Can we discuss it before closing.

I agree that it is extended validation for IBAN ... but how we can process such issues in user application?

Should we check IBAN isself and next BBAN .... by own code ...
It will be additional implemetaton needed in user application.
In such case we user can implent everythig in own way. So we will not need external library.

I hope that such extension will be usefully for users - we simply check if account is valid or not.

garydgregory:
Is there a compromise where the code can be refactored to make it easier for a user to extend the behavior either through a subclass or a lamda parameter?

slawekjaranowski:
I'm not sure if it is good way to provide new api that each user should extends to validate. 
For such case CZ IBAN account every user need to extend in similar way ... for me looks like do it in one place will bring more advantages.

You can try with example account `CZ8801007822637402170100` on some other IBAN validators like:
https://www.ibancalculator.com/iban_validieren.html

slawekjaranowski:
> > You can try with example account `CZ8801007822637402170100` on some other IBAN validators like: https://www.ibancalculator.com/iban_validieren.html
> 
> However, they do not support BBAN validation for all countries, even though they offer a paid service:

So we can start from simple one ... If somebody find useful for next country it can be added in the future.

> 
> https://www.ibancalculator.com/data_quality.html
> 
> That appears to be because of the difficulty and/or expense of getting the required information to do the checks.
> 
> Or, in the case of Greek codes, the pattern is too irregular: https://www.ibancalculator.com/guarantee.html

PR #379, opened 2026-05-30
Reject signed hex group in InetAddressValidator.isValidInet6Address(String)

sahvx655-wq:
isValidInet6Address parses each hex group with Integer.parseInt(octet, 16), which tolerates a leading '+', so addresses like 1:2:3:4:5:6:7:+8, fe80::+1 and ::+f are accepted even though they are not valid IPv6 literals (JDK InetAddress.getByName rejects them); the '-' form is already caught by the existing octetInt < 0 check, so only '+' slips through.


garydgregory:
@sahvx655-wq 

Thank you for the PR. I merged the PR and followed up with a clean up to remove an unnecessary duplicate call.

PR #380, opened 2026-05-31
reject non-ASCII Unicode digits in check-digit routines

sahvx655-wq:
Character.getNumericValue/isDigit accept fullwidth and Arabic-Indic digits, so Unicode-digit codes pass isValid on EAN-13, Luhn, ISIN, CUSIP, SEDOL and Verhoeff; restrict each conversion site to ASCII.


garydgregory:
Closing: Build fails, no follow up, provide a better implementation in https://github.com/apache/commons-validator/pull/385. You are credited in `changes.xml`.

PR #381, opened 2026-06-03
Fix array serialization handling in ValidatorUtils.getValueAsString()

sahvx655-wq:
This PR fixes incorrect serialization behavior in `ValidatorUtils.getValueAsString()` for `String[]` values. Previously, arrays were converted using the default `Object.toString()` implementation, resulting in non-readable JVM identity strings.
When a `String[]` was passed to `getValueAsString()`, the output appeared as:
text
[Ljava.lang.String;@1a2b3c

This occurs because Java arrays do not override `toString()`.

Updated array handling to use:
java
java.util.Arrays.toString((String[]) value)

This produces consistent and human-readable output:
text
[element1, element2, element3]

### DIFF
--- a/src/main/java/org/apache/commons/validator/ValidatorResources.java
+++ b/src/main/java/org/apache/commons/validator/ValidatorResources.java
@@ -31,6 +31,8 @@ import org.apache.commons.digester.xmlrules.DigesterLoader;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.xml.sax.Attributes;
+import org.xml.sax.EntityResolver;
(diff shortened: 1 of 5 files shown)


garydgregory:
@sahvx655-wq 

-1: There are 2 major problems with this PR.

You or your AI are hallucinating requirements and the test is absurdly complicated.

The Javadoc for the method you edited reads "Otherwise, property.toString() is returned.". So this PR not only breaks the behavior that has been explicitly documented for 20+ years, it ALSO doesn't update the Javadoc.

The test is absurd. You are changing a public API, so you can just have the test call the public API without all this useless XML scaffolding.

PR #382, opened 2026-06-03
Fix state leakage of otherCountryCodes in IBANValidator custom configurations

sahvx655-wq:
This PR fixes a state consistency issue in mutable `IBANValidator` instances where associated territory codes (`otherCountryCodes`) could remain mapped to obsolete validator instances after a validator was removed or replaced.

Some validators define additional territory codes through `otherCountryCodes` (for example, `FR` includes `GF`, `GP`, `MQ`, etc., and `GB` includes `IM`, `JE`, `GG`).

When `setValidator(...)` was used to remove or replace a validator, the implementation updated only the primary `countryCode` entry in `validatorMap`. Any entries created for `otherCountryCodes` were left untouched.

As a result, territory codes could continue resolving to a stale validator that was no longer registered, causing inconsistent validation behavior and leaving `validatorMap` in an incorrect state.

The validator update logic has been extended to keep all related mappings synchronized:

- When a validator is removed, any associated `otherCountryCodes` mappings are removed as well.
- When a validator is replaced, mappings belonging to the previous validator are cleaned up before registering the new validator and its associated territory codes.
- The internal `validatorMap` now remains consistent after validator additions, replacements, and removals.
Added regression tests in `IBANValidatorTest` covering:

- Removal of a validator with associated `otherCountryCodes`
- Replacement of a validator with associated `otherCountryCodes`
- Verification that stale territory-code mappings are properly cleaned up
- Verification that new territory-code mappings are correctly registered

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/IBANValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/IBANValidator.java
@@ -376,7 +376,13 @@ public class IBANValidator {
             throw new IllegalStateException("The singleton validator cannot be modified");
         }
         if (length < 0) {
-            return validatorMap.remove(countryCode);
+            final Validator prev = validatorMap.remove(countryCode);
+            if (prev != null) {
+                for (final String otherCC : prev.otherCountryCodes) {
+                    validatorMap.remove(otherCC);
+                }
+            }
+            return prev;
         }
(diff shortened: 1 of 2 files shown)


garydgregory:
Closing: Broken build, fixed in https://github.com/apache/commons-validator/pull/386. You are credited in `changes.xml`.

PR #383, opened 2026-06-03
Prevent URL path traversal bypass via percent encoding in UrlValidator

sahvx655-wq:
### Summary

This PR fixes a path validation bypass in `UrlValidator` where percent-encoded path traversal sequences could evade the existing parent-directory checks performed by `isValidPath(String)`.

`UrlValidator` validates paths by normalizing the URL path and rejecting traversal attempts such as `/../`.

However, the validation logic operates on the raw path component. Percent-encoded traversal sequences (for example `%2e%2e` and `%2f`) are not interpreted during normalization and therefore are not collapsed into their equivalent `../` path segments.

As a result, inputs such as:

http://example.com/..%2fetc/passwd
http://example.com/%2e%2e/world

can bypass the existing traversal checks and be considered valid.

Decode traversal-relevant percent-encoded characters before normalization so that encoded traversal sequences are evaluated consistently with their literal equivalents.

This ensures that both literal and percent-encoded parent-directory traversal attempts are rejected by the existing validation logic.

Added regression tests covering:

* `..%2f`
* `..%2F`
* `%2e%2e/`
* `%2e%2e%2f`
* mixed-case encoded traversal sequences

The new tests fail prior to the fix and pass after the fix.

This change closes a validation gap in `UrlValidator` and ensures that directory traversal checks cannot be bypassed using percent-encoded path separators or dot segments.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
@@ -474,6 +474,32 @@ public class UrlValidator implements Serializable {
         return isOff(NO_FRAGMENTS);
     }
 
+    private String decodePath(final String path) {
+        final StringBuilder sb = new StringBuilder();
+        final int len = path.length();
+        for (int i = 0; i < len; i++) {
+            final char c = path.charAt(i);
+            if (c == '%' && i + 2 < len) {
+                final char h1 = path.charAt(i + 1);
+                final char h2 = path.charAt(i + 2);
+                if (h1 == '2') {
+                    if (h2 == 'e' || h2 == 'E') {
+                        sb.append('.');
+                        i += 2;
+                        continue;
+                    }
+                    if (h2 == 'f' || h2 == 'F') {
+                        sb.append('/');
+                        i += 2;
+                        continue;
+                    }
+                }
+            }
+            sb.append(c);
+        }
+        return sb.toString();
+    }
+
     /**
      * Returns true if the path is valid.  A {@code null} value is considered invalid.
      *
@@ -487,7 +513,7 @@ public class UrlValidator implements Serializable {
 
         try {
             // Don't omit host otherwise leading path may be taken as host if it starts with //
-            final URI uri = new URI(null, "localhost", path, null);
+            final URI uri = new URI(null, "localhost", decodePath(path), null);
             final String norm = uri.normalize().getPath();
             if (norm.startsWith("/../") // Trying to go via the parent dir
                     || norm.equals("/..")) { // Trying to go to the parent dir
--- a/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
@@ -489,10 +489,7 @@ public class UrlValidator implements Serializable {
         }
 
         try {
-            // URLDecoder.decode converts '+' to a space character.
-            // Since '+' is a valid literal character in a URI path,
-            // we escape it to '%2B' before decoding to preserve it.
-            final String decodedPath = URLDecoder.decode(path.replace("+", "%2B"), StandardCharsets.UTF_8.name());
+            final String decodedPath = URLDecoder.decode(path, StandardCharsets.UTF_8.name());
             // Don't omit host otherwise leading path may be taken as host if it starts with //
             final URI uri = new URI(null, "localhost", decodedPath, null);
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/validator/routines/UrlValidator.java
```diff
@@ -487,7 +513,7 @@ protected boolean isValidPath(final String path) {
 
         try {
             // Don't omit host otherwise leading path may be taken as host if it starts with //
-            final URI uri = new URI(null, "localhost", path, null);
+            final URI uri = new URI(null, "localhost", decodePath(path), null);
```
sebbASF: Why not use URLDecoder directly on the path?

It does not seem right to provide a home-built percent decoder here.

sebbASF (changes requested):

sahvx655-wq:
Thanks for the feedback. I've updated the PR to use `URLDecoder.decode` instead of the custom logic. Literal `+` characters are preserved, invalid encodings fall back to the original path,

src/main/java/org/apache/commons/validator/routines/UrlValidator.java
```diff
@@ -474,6 +477,17 @@ protected boolean isValidFragment(final String fragment) {
         return isOff(NO_FRAGMENTS);
     }
 
+    private String decodePath(final String path) {
+        try {
+            // URLDecoder.decode converts '+' to a space character.
+            // Since '+' is a valid literal character in a URI path,
+            // we escape it to '%2B' before decoding to preserve it.
```
sebbASF: This is unnecessary

src/main/java/org/apache/commons/validator/routines/UrlValidator.java
```diff
@@ -474,6 +477,17 @@ protected boolean isValidFragment(final String fragment) {
...
+            return URLDecoder.decode(path.replace("+", "%2B"), StandardCharsets.UTF_8.name());
+        } catch (final UnsupportedEncodingException | IllegalArgumentException e) {
+            return path;
```
sebbASF: If the path cannot be decoded, it should be treated as invalid, same as is done for URISyntaxException in the method isValidPath

sebbASF (changes requested):

src/main/java/org/apache/commons/validator/routines/UrlValidator.java
```diff
@@ -476,6 +479,14 @@ protected boolean isValidFragment(final String fragment) {
         return isOff(NO_FRAGMENTS);
     }
 
+    private String decodePath(final String path) {
```
sebbASF: Not sure there is any need to use a private method here.
It would be simpler to just catch UnsupportedEncodingException in isvalid

src/main/java/org/apache/commons/validator/routines/UrlValidator.java
```diff
@@ -476,6 +479,14 @@ protected boolean isValidFragment(final String fragment) {
...
+        try {
+            return URLDecoder.decode(path, StandardCharsets.UTF_8.name());
+        } catch (final UnsupportedEncodingException e) {
+            throw new RuntimeException(e); // UTF-8 is always supported
```
garydgregory: Throwing RE is an anti-pattern, IAE is better here IMO. There's also UncheckedIOException if the IO aspect is more important.

sebbASF: Given that the exception is immediately caught by the caller, there is no need to rethrow.

garydgregory:
@sahvx655-wq 
Ping.

sahvx655-wq:
@garydgregory,
Thanks for the ping.
I'm working through the remaining review comments from @sebbASF and will push an updated version shortly

sahvx655-wq:
I have updated the PR based on the review feedback. The path decoding logic is now handled directly in isValidPath(), decoding/parsing failures are treated as validation failures, and the consecutive slash check is performed on the decoded path. I have also added regression tests covering the reported traversal and encoding scenarios.
I have pushed the changes for review

sebbASF:
Please correct build failure

sahvx655-wq:
I have fixed the build failure and pushed the update

sebbASF:
Thanks for your patience in getting this resolved.

garydgregory:
@sahvx655-wq 
Yes, thank you!

PR #384, opened 2026-06-03
reject NaN and infinity in double and float validators

sahvx655-wq:
reading the number validators, validate of double and float passes the parsed value straight through processParsedValue without checking it is finite. NumberFormat recognises the locale NaN and infinity symbols, so validate("NaN"), validate("∞") and validate("-∞") return NaN/infinity and isValid returns true. a NaN that validates then silently defeats every isInRange/minValue/maxValue check, since NaN compares false against everything. reject non-finite results in both validators.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/DoubleValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/DoubleValidator.java
@@ -193,10 +193,16 @@ public class DoubleValidator extends AbstractNumberValidator {
      */
     @Override
     protected Object processParsedValue(final Object value, final Format formatter) {
+        final double doubleValue = ((Number) value).doubleValue();
+        // The "Infinity"/"-Infinity" strings throw a ParseException, but the locale infinity
+        // symbol parses through to an infinite value; reject it so the two paths are consistent.
+        if (Double.isInfinite(doubleValue)) {
+            return null;
+        }
         if (value instanceof Double) {
             return value;
         }
-        return Double.valueOf(((Number) value).doubleValue());
+        return Double.valueOf(doubleValue);
 
     }
 
--- a/src/main/java/org/apache/commons/validator/routines/FloatValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/FloatValidator.java
@@ -197,6 +197,11 @@ public class FloatValidator extends AbstractNumberValidator {
 
         final double doubleValue = ((Number) value).doubleValue();
 
+        // The "Infinity"/"-Infinity" strings throw a ParseException, but the locale infinity
+        // symbol parses through to an infinite value; reject it so the two paths are consistent.
+        if (Double.isInfinite(doubleValue)) {
+            return null;
+        }
         if (doubleValue > 0) {
             if (doubleValue < Float.MIN_VALUE || doubleValue > Float.MAX_VALUE) {
                 return null;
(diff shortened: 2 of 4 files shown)


garydgregory:
@sahvx655-wq 

Did you read the Javadoc?

This looks like you or your AI are imagining requirements that don't exist or that I can't find.

For example, "NaN" is a valid value to parse with a `NumberFormat` (which we use):
```
jshell> import java.text.*;
jshell> NumberFormat.getInstance().parse("NaN");
$3 ==> NaN
```
But it's not the same for `Double.POSITIVE_INFINITY`:
```
jshell> NumberFormat.getInstance().parse("" + Double.POSITIVE_INFINITY)
|  Exception java.text.ParseException: Unparseable number: "+Infinity"
|        at NumberFormat.parse (NumberFormat.java:434)
|        at (#2:1)
```
So there is a discrepancy...

Aside from the issue that the PR this may break existing applications, how do you propose an app use this class to validate, for example, config file values, or user input where in fact, a Double value like `Double.NaN` is expected?

FYI, I updated tests for edge case tests in `FloatValidatorTest` and `DoubleValidatorTest` to reflect the current code.

TY!

sahvx655-wq:
fair point. NaN is a legitimate Double value and NumberFormat parses it by design, so rejecting it here is a behaviour change that could break callers who genuinely want to validate it. my reasoning was that a NaN which passes then slips past every isInRange/minValue check, since NaN compares false against everything, but on reflection that guard belongs to the caller, not this class. and as you spotted the infinity handling is already inconsistent (NaN and the ∞ symbol parse, +Infinity doesn't), which isn't something to paper over in this PR. happy to close this out, your edge-case test updates already pin the current behaviour as the sensible default.

garydgregory:
Also note that NaN parsing behavior is different on Java 8 compared to newer versions.

garydgregory:
Do you think it is worth fixing the `Double.[POSITIVE|NEGATIVE]_INFINITY` cases?

sahvx655-wq:
Yes, I think the infinity cases are worth closing. I traced the parse with a ParsePosition under Locale.US: the "Infinity"/"-Infinity" strings throw a ParseException so validate already returns null, but the locale infinity symbol (the U+221E glyph) parses with full consumption straight to POSITIVE_INFINITY, so the strict check passes and validate hands back an infinite Double. Same value, two spellings, opposite outcomes, which is the discrepancy you pointed at.

So I've narrowed this PR to reject only infinite results in both processParsedValue methods and left NaN exactly as your edge-case tests pin it, since that one is symmetric and a legitimate value to validate. The risk if left is the same shape as the NaN one: an infinite value sails past isInRange/maxValue because the comparisons don't bound it, but here there's no matching "Infinity" string that validates, so nothing legitimate relies on the symbol getting through. Added a test in each class that derives the symbol from DecimalFormatSymbols so it follows the locale, and no JRE guard is needed because on Java 8 the symbol just fails to parse and still returns null.

garydgregory:
Somehow, you or your AI tool the opposite of my meaning. Instead of this frustrating loop, I created https://github.com/apache/commons-validator/pull/388 and will close this PR.

sahvx655-wq:
ah I get it now, you wanted them accepted so they're all consistent, not rejected. I read it backwards. thanks for sorting it in #388. still pretty new to this stuff but learning, appreciate the patience.

garydgregory:
YW. Your help is greatly appreciated 👏

PR #387, opened 2026-06-05
Fix long range check in GenericTypeValidator.formatLong

sahvx655-wq:
While going over the numeric format helpers I noticed formatLong(String, Locale) range-checks the parsed number with num.doubleValue() against Long.MIN_VALUE and Long.MAX_VALUE. Long.MAX_VALUE is not exactly representable as a double, so it is promoted to 2^63 in that comparison. Parsing "9223372036854775808" (Long.MAX_VALUE + 1), which NumberFormat returns as a Double that rounds to 2^63, then satisfies the upper bound and is accepted; longValue() afterwards clamps it back to Long.MAX_VALUE, so the helper reports a value the caller never entered. "-9223372036854775809" behaves the same way at the lower bound. The root cause is using a lossy double comparison to decide whether an integral value fits in a long.

The fix accepts the result only when NumberFormat hands back a Long, which it does exactly when the value fits in a long; anything out of range comes back as a Double and is now rejected. This matches what routines.LongValidator already does and keeps the bounds decision inside the conversion routine instead of leaving callers to second-guess a silently clamped result. Left unfixed this lets out-of-range input pass validation and corrupt the converted value. The other format* methods are untouched because their int/short/byte bounds are exact as doubles, so the same rounding gap does not arise.

### DIFF
--- a/src/main/java/org/apache/commons/validator/GenericTypeValidator.java
+++ b/src/main/java/org/apache/commons/validator/GenericTypeValidator.java
@@ -405,10 +405,13 @@ public class GenericTypeValidator implements Serializable {
             final Number num = formatter.parse(value, pos);
 
             // If there was no error      and we used the whole string
+            // NumberFormat returns a Long only when the value fits in a long;
+            // out-of-range input is returned as a Double, so a doubleValue()
+            // range check cannot be used here (Long.MAX_VALUE is not exactly
+            // representable as a double and would let 2^63 through).
             if (pos.getErrorIndex() == -1 && pos.getIndex() == value.length() &&
-                    num.doubleValue() >= Long.MIN_VALUE &&
-                    num.doubleValue() <= Long.MAX_VALUE) {
-                result = Long.valueOf(num.longValue());
+                    num instanceof Long) {
+                result = (Long) num;
             }
         }
 
--- a/src/main/java/org/apache/commons/validator/GenericTypeValidator.java
+++ b/src/main/java/org/apache/commons/validator/GenericTypeValidator.java
@@ -404,13 +404,14 @@ public class GenericTypeValidator implements Serializable {
             final ParsePosition pos = new ParsePosition(0);
             final Number num = formatter.parse(value, pos);
 
-            // If there was no error      and we used the whole string
+            // If we used the whole string and the value fits in a long.
             // NumberFormat returns a Long only when the value fits in a long;
             // out-of-range input is returned as a Double, so a doubleValue()
             // range check cannot be used here (Long.MAX_VALUE is not exactly
-            // representable as a double and would let 2^63 through).
-            if (pos.getErrorIndex() == -1 && pos.getIndex() == value.length() &&
-                    num instanceof Long) {
+            // representable as a double and would let 2^63 through). A failed
+            // parse returns null, so the instanceof check also covers it and
+            // the pos.getErrorIndex() test is no longer needed.
+            if (pos.getIndex() == value.length() && num instanceof Long) {
                 result = (Long) num;
             }
         }
--- a/src/test/java/org/apache/commons/validator/GenericTypeValidatorTest.java
+++ b/src/test/java/org/apache/commons/validator/GenericTypeValidatorTest.java
@@ -22,6 +22,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 
 import java.io.IOException;
+import java.math.BigInteger;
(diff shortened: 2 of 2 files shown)


src/test/java/org/apache/commons/validator/GenericTypeValidatorTest.java
```diff
@@ -171,6 +172,18 @@ void testType() throws ValidatorException {
 
     }
 
+    /**
+     * Tests that {@link GenericTypeValidator#formatLong(String, Locale)} rejects values just outside the long range instead of clamping them.
+     */
+    @Test
+    void testLongLocaleOverflow() {
+        assertEquals(Long.valueOf(Long.MAX_VALUE), GenericTypeValidator.formatLong("9223372036854775807", Locale.US));
```
garydgregory: Don't use magic strings here. Use Long.MAX|MIN_VALUE converted to a String.

garydgregory (changes requested):
Hello @sahvx655-wq 
Thank you for your PR.
Please see my comment.

garydgregory:
@sahvx655-wq 
ping.

src/test/java/org/apache/commons/validator/GenericTypeValidatorTest.java
sahvx655-wq: Good point. Pushed a commit that drops the literals: the in-range cases now use Long.toString(Long.MAX_VALUE)/Long.toString(Long.MIN_VALUE), and the two overflow cases derive MAX+1 and MIN-1 via BigInteger so nothing is hand-typed.

Confirmed the derived strings match the old literals exactly and the assertions still hold (bounds accepted, the two out-of-range values rejected).

src/main/java/org/apache/commons/validator/GenericTypeValidator.java
```diff
@@ -405,10 +405,13 @@ public static Long formatLong(final String value, final Locale locale) {
             final Number num = formatter.parse(value, pos);
 
             // If there was no error      and we used the whole string
+            // NumberFormat returns a Long only when the value fits in a long;
+            // out-of-range input is returned as a Double, so a doubleValue()
+            // range check cannot be used here (Long.MAX_VALUE is not exactly
+            // representable as a double and would let 2^63 through).
```
garydgregory: Hi @sahvx655-wq 

Thank you for your update.

Either we have tests missing or checking `pos` for errors is superfluous. WDYT?

sahvx655-wq: You're right on one half and it's worth splitting them apart. The two pos conditions do different jobs. pos.getErrorIndex() == -1 only matters when the parse fails outright, but in that case NumberFormat.parse returns null, and num instanceof Long is already false for null, so that test was dead weight once the range check became an instanceof. I've dropped it.

The pos.getIndex() == value.length() one isn't superfluous though: with setParseIntegerOnly(true) a value like "123x" parses to a Long 123 and stops at the first non-digit, leaving getIndex() at 3 against a length of 4, so without the whole-string check that trailing rubbish would be accepted. That path had no coverage, which is why it read as redundant, so I've added an assertNull for "123x" alongside the overflow cases. Patched tree accepts the bounds and rejects MAX+1, MIN-1 and "123x".

garydgregory:
TY @sahvx655-wq 
Merged 🚀

garydgregory:
@sahvx655-wq 

Do you see any of the other conversions suffering from similar overflow (or underflow) issues?

TY!

sahvx655-wq:
I went back through the rest of GenericTypeValidator after this one. Short answer is no, formatLong was the only one with the gap. formatByte, formatShort and formatInt all range-check with num.doubleValue() the same way, but their bounds (±127, ±32767, ±2^31-1) sit well under 2^53, so they're exactly representable as doubles and the comparison is exact. Anything too large to be a long comes back from NumberFormat as a Double, and it's still larger than the int/short/byte maximum, so it gets rejected rather than clamped.

formatDouble and formatFloat compare against Double.MAX_VALUE and Float.MAX_VALUE, and those constants are themselves exact as doubles, so nothing above the type maximum slips through. The float path still narrows in-range values with the usual precision loss, but that's the documented behaviour, not an out-of-range value being accepted and silently changed. The long case was the odd one out precisely because Long.MAX_VALUE (2^63-1) is the single bound that isn't representable as a double and rounds up to 2^63, which is what opened the hole. So I don't think there's a follow-up needed on the others.

garydgregory:
@sahvx655-wq 

Thank you for your analysis 💯

PR #394, opened 2026-06-08
fix BigIntegerValidator clamping values above Long.MAX_VALUE

sahvx655-wq:
BigIntegerValidator.validate runs the input through NumberFormat, which returns a Long while the number fits in a long and switches to a Double once it overflows. processParsedValue then converts via longValue(), and on a Double that saturates at Long.MAX_VALUE, so "99999999999999999999999999" comes back as 9223372036854775807 with no exception or warning to flag it. I caught this lining the integer validators up against each other: BigDecimalValidator routes the same parsed Double through its string form and keeps the magnitude, while BigIntegerValidator throws it away.

Sending the non-Long branch through the parsed value's string form mirrors what BigDecimalValidator already does, so the two stay consistent on oversized input. Keeping the conversion inside processParsedValue covers every validate and isValid entry point without touching the shared parsing code. Left alone the behaviour is a quiet correctness hole: folding every large value onto one fixed number lets a later maxValue or range check pass for inputs the real number should fail.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
@@ -158,8 +158,6 @@ public class BigIntegerValidator extends AbstractNumberValidator {
      */
     @Override
     protected Object processParsedValue(final Object value, final Format formatter) {
-        // A value that fits in a long is parsed as a Long; anything larger comes back as a
-        // Double, whose longValue() saturates at Long.MAX_VALUE and loses the magnitude.
         if (value instanceof Long) {
             return BigInteger.valueOf(((Long) value).longValue());
         }
--- a/src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
@@ -16,6 +16,7 @@
  */
 package org.apache.commons.validator.routines;
 
+import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.text.Format;
 import java.text.NumberFormat;
@@ -157,7 +158,12 @@ public class BigIntegerValidator extends AbstractNumberValidator {
      */
     @Override
     protected Object processParsedValue(final Object value, final Format formatter) {
-        return BigInteger.valueOf(((Number) value).longValue());
+        // A value that fits in a long is parsed as a Long; anything larger comes back as a
+        // Double, whose longValue() saturates at Long.MAX_VALUE and loses the magnitude.
+        if (value instanceof Long) {
+            return BigInteger.valueOf(((Long) value).longValue());
+        }
+        return new BigDecimal(value.toString()).toBigInteger();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Please don't use magic strings in the tests. Instead use Long max value to make sure we are really above it.

sahvx655-wq:
Good catch. I've dropped the literal strings and built the input from `BigInteger.valueOf(Long.MAX_VALUE).add(BigInteger.ONE)`, so it's the first value that can't fit in a long and is provably above the boundary rather than a hand-counted digit string.

The two assertions stand on their own now: the result has to compare greater than Long.MAX_VALUE (the old code clamped it exactly there), and it has to agree with what BigDecimalValidator returns for the same input. I reran it against the unpatched processParsedValue to confirm it still fails there and passes after the fix.

src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
```diff
@@ -157,7 +158,12 @@ public boolean minValue(final BigInteger value, final long min) {
      */
     @Override
     protected Object processParsedValue(final Object value, final Format formatter) {
-        return BigInteger.valueOf(((Number) value).longValue());
+        // A value that fits in a long is parsed as a Long; anything larger comes back as a
+        // Double, whose longValue() saturates at Long.MAX_VALUE and loses the magnitude.
```
sebbASF: The comment does not make sense to me

src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
```diff
@@ -157,7 +158,12 @@ public boolean minValue(final BigInteger value, final long min) {
...
+        if (value instanceof Long) {
+            return BigInteger.valueOf(((Long) value).longValue());
+        }
+        return new BigDecimal(value.toString()).toBigInteger();
```
sebbASF, line 164: According to the Javadoc, processParsedValue always returns a BigInteger

sebbASF (changes requested):

src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
sahvx655-wq: Fair point, that comment was describing the old failure path rather than the code in front of you, so it just read as noise. I've removed it in 2e71709. The two branches now mirror BigDecimalValidator.processParsedValue, which has the same Long-versus-string split and carries no comment.

src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
sahvx655-wq: Right, both branches return a BigInteger: the Long case through BigInteger.valueOf, anything larger through new BigDecimal(value.toString()).toBigInteger(). The split is only about how the value arrives, not the type that comes back. NumberFormat returns a Long while the input still fits in a long and switches to a Double once it overflows, and Double.longValue() saturates at Long.MAX_VALUE, which is what was quietly dropping the magnitude. Routing the non-Long case through its string form keeps the digits and lines up with what BigDecimalValidator already does.

garydgregory: @sebbASF 
Does @sahvx655-wq 's reply allay your concern?

garydgregory:
Co-pilot generated code that doesn't compile; fixing...

garydgregory:
@sahvx655-wq 
I see now that PR is only a partial fix for what's wrong with this class. I'll create a new PR...

garydgregory:
@sahvx655-wq 
Closing in favor of https://github.com/apache/commons-validator/pull/397. You'll be credited in `changes.xml`.

PR #399, opened 2026-06-15
Fix ISIN country-code check using untrimmed input

sahvx655-wq:
**ISIN country code read from the raw argument**

Whilst checking the country-code path I noticed that `isValid`/`validate` disagree with the underlying `CodeValidator` for inputs with surrounding whitespace. The validator trims the argument before matching, so `validate(" US0378331005")` returns the trimmed `US0378331005`, but the ISO-3166 lookup then takes `code.substring(0, 2)` from the original untrimmed string and reads `" U"`, which is not a country code. The upshot is that the same valid ISIN passes with `getInstance(false)` yet is rejected by `getInstance(true)`.

The root cause is the country code being derived from the raw argument rather than from the code the validator actually accepted. Reading it from the validated value keeps both code paths consistent. Left unfixed, callers that enable the country check silently reject otherwise-valid codes purely on the presence of leading whitespace. Added a regression test for the trimmed lookup.

### DIFF
--- a/src/test/java/org/apache/commons/validator/routines/ISINValidatorTest.java
+++ b/src/test/java/org/apache/commons/validator/routines/ISINValidatorTest.java
@@ -16,6 +16,7 @@
  */
 package org.apache.commons.validator.routines;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
@@ -108,6 +109,17 @@ class ISINValidatorTest {
         for (final String f : validFormat) {
             assertTrue(VALIDATOR_TRUE.isValid(f), f);
         }
+    }
+
+    @Test
+    void testValidWithSurroundingWhitespaceCountryCode() {
+        // The underlying CodeValidator trims the input, so the country code must be
+        // derived from the trimmed code. Otherwise a valid ISIN with leading whitespace
+        // is accepted without the country check but rejected with it.
+        final String leading = " US0378331005";
+        assertTrue(VALIDATOR_FALSE.isValid(leading), leading);
+        assertTrue(VALIDATOR_TRUE.isValid(leading), leading);
+        assertEquals("US0378331005", VALIDATOR_TRUE.validate(leading));
     }
 
 }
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/validator/routines/ISINValidatorTest.java
```diff
@@ -110,4 +111,15 @@ void testIsValidTrue() {
         }
     }
 
+    @Test
+    void testValidWithSurroundingWhitespaceCountryCode() {
+        // The underlying CodeValidator trims the input, so the country code must be
+        // derived from the trimmed code. Otherwise a valid ISIN with leading whitespace
+        // is accepted without the country check but rejected with it.
+        final String leading = " US0378331005";
```
garydgregory: How about trailing whitespace?

garydgregory (changes requested):
Hello @sahvx655-wq 

Thank you for the PR.

Please see my comment.

src/test/java/org/apache/commons/validator/routines/ISINValidatorTest.java
sahvx655-wq: Trailing whitespace is handled the same way. CodeValidator.validate trims both ends before matching, and since the fix now reads the country code from the validated/trimmed value rather than the raw argument, the substring(0, 2) lookup sees "US" regardless of where the whitespace sits.

I have extended the regression test to loop over leading, trailing and surrounding whitespace so all three paths are covered. Full ISINValidatorTest still passes (Tests run: 5, Failures: 0).

src/test/java/org/apache/commons/validator/routines/ISINValidatorTest.java
```diff
@@ -110,4 +111,16 @@ void testIsValidTrue() {
         }
     }
 
+    @Test
+    void testValidWithSurroundingWhitespaceCountryCode() {
+        // The underlying CodeValidator trims the input, so the country code must be
+        // derived from the trimmed code. Otherwise a valid ISIN with surrounding whitespace
+        // is accepted without the country check but rejected with it.
+        for (final String code : new String[] { " US0378331005", "US0378331005 ", " US0378331005 " }) {
```
garydgregory: Use `@ParameterizedTest` instead of a loop.

garydgregory (changes requested):
@sahvx655-wq 
Please see my comment.
TY.

src/test/java/org/apache/commons/validator/routines/ISINValidatorTest.java
sahvx655-wq: Done. Swapped the loop for a @ParameterizedTest with @ValueSource covering the leading, trailing and surrounding whitespace cases. Each case now runs as its own invocation, so a failure points straight at the offending input rather than aborting the whole loop. Full ISINValidatorTest is green (7 tests, 0 failures).

PR #400, opened 2026-06-19
Use compareTo() in BigIntegerValidator.minValue()

sahvx655-wq:
minValue narrows the BigInteger to a long before comparing, so any value outside long range gives the wrong answer, whereas the sibling maxValue and isInRange already compare with compareTo.

1. value.longValue() keeps only the low 64 bits, so a value such as 2^63 is compared as 0 and a value just past Long.MIN_VALUE wraps to a large positive, which makes minValue disagree with this class's own maxValue/isInRange.
2. commit eafeda9 moved isInRange and maxValue onto compareTo(BigInteger.valueOf(...)) but left minValue on longValue(), so the gap stayed: minValue(2^64, 5) returns false although 2^64 is clearly greater than 5.

I switched minValue to the same compareTo form. Two assertions in testBigIntegerAboveLongMaxValue and testBigIntegerBelowLongMinValue were matching the truncated output (a value above Long.MAX_VALUE was treated as not greater-or-equal to Long.MAX_VALUE), so I corrected those to the right results and added a small test built from exact BigIntegers, which sidesteps the double rounding the parsed cases go through.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
@@ -145,7 +145,7 @@ public class BigIntegerValidator extends AbstractNumberValidator {
      *         or equal to the minimum.
      */
     public boolean minValue(final BigInteger value, final long min) {
-        return value.longValue() >= min;
+        return value.compareTo(BigInteger.valueOf(min)) >= 0;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


garydgregory:
Merged 🚀 Thank you @sahvx655-wq 
Please review other BigInteger and BigDecimal to primitive conversion issues if you can.
TY!

PR #401, opened 2026-06-19
Compare Exact value in BigDecimalValidator min/max range checks

sahvx655-wq:
**Range checks on BigDecimalValidator narrow the value to a double before comparing**

`minValue`, `maxValue` and `isInRange` ran the supplied `BigDecimal` through `doubleValue()` and compared the resulting primitive against the bound. Any value that only differs from the bound past the double mantissa is rounded onto the bound first, so the comparison is settled on a value the caller never passed. Following up on the `BigIntegerValidator` work I checked the same range checks here: `maxValue(2^53 + 1, 2^53)` returns `true`, because `2^53 + 1` has no exact double and rounds back to `2^53`, and `minValue` misfires the same way for a value that rounds up onto its minimum. For a validation routine that is the failure that matters, since it lets an out-of-range value pass a bound check.

The comparison now runs against the exact `BigDecimal` with `compareTo(BigDecimal.valueOf(bound))` for finite bounds, the same way `BigIntegerValidator` already compares. `BigDecimal.valueOf` rejects NaN and infinity, so non-finite bounds keep the `doubleValue()` path and the `±Infinity` behaviour covered by `testBigDecimalBeyondDoubleRange` is unchanged. `isInRange` delegates to the two methods so the rule lives in one place. The added test fails on the current code and passes with the patch.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/BigDecimalValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/BigDecimalValidator.java
@@ -135,7 +135,7 @@ public class BigDecimalValidator extends AbstractNumberValidator {
      *         specified range.
      */
     public boolean isInRange(final BigDecimal value, final double min, final double max) {
-        return value.doubleValue() >= min && value.doubleValue() <= max;
+        return minValue(value, min) && maxValue(value, max);
     }
 
     /**
@@ -147,6 +147,9 @@ public class BigDecimalValidator extends AbstractNumberValidator {
      *         or equal to the maximum.
      */
     public boolean maxValue(final BigDecimal value, final double max) {
+        if (Double.isFinite(max)) {
+            return value.compareTo(BigDecimal.valueOf(max)) <= 0;
+        }
         return value.doubleValue() <= max;
     }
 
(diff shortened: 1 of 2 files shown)


garydgregory:
@sahvx655-wq 
Please also add tests (and possibly main changes) that cover:
- `Double.NaN`
- `Double.NEGATIVE_INFINITY`
- `Double.POSITIVE_INFINITY`

sahvx655-wq:
Added `testBigDecimalNonFiniteBounds` covering all three as bounds for `minValue`, `maxValue` and `isInRange`. No main change was needed: `BigDecimal.valueOf` rejects non-finite doubles, so a `NaN` or `±Infinity` bound keeps the `doubleValue()` comparison path, which already behaves sensibly. A `NaN` bound is never satisfied because every comparison against `NaN` is false; `POSITIVE_INFINITY` as a maximum (and `NEGATIVE_INFINITY` as a minimum) is an open bound any finite value meets, while the reverse pair cannot be met by a finite value.

The new test asserts each of those cases. Full suite stays green (1068 tests) and the default `mvn` goal builds clean.

garydgregory:
Merged 🚀 Thank you @sahvx655-wq

PR #405, opened 2026-06-20
Accept IPv4-embedded IPv6 addresses in UrlValidator

sahvx655-wq:
While checking how UrlValidator handles IPv6 literals I noticed it accepts `http://[::FFFF:129.144.52.38]/` (the case covered by `testValidator452`) but rejects the same address in the standard lower-case mapped form `http://[::ffff:129.144.52.38]/`, and likewise rejects the IPv4-compatible `::1.2.3.4` and embedded `2001:db8::1.2.3.4` notations. The cause is `IPV6_REGEX` used by `isValidAuthority`: it only special-cases the literal upper-case `::FFFF:` prefix and otherwise matches `[0-9a-fA-F:]+`, which has no `.`, so any bracketed host with an embedded IPv4 part fails the authority match before `InetAddressValidator.isValidInet6Address` is ever consulted. That validator already accepts all of these forms, so the URL path and a direct address check disagree.

The bracketed group is captured and then handed to `isValidInet6Address`, so the regex only needs to delimit the candidate rather than judge it; I widened it to `[0-9a-fA-F:.]+` and left the real decision to the address validator. Keeping the judgement in one place stops the two paths drifting apart again, and malformed hosts (bad octet, stray `::`, over-long group) are still rejected because the address validator rejects them. Left unfixed, code using UrlValidator as an allow gate silently turns away legitimate IPv4-mapped IPv6 URLs.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
@@ -118,8 +118,9 @@ public class UrlValidator implements Serializable {
     // TODO does not allow for optional userinfo.
     // Validation of character set is done by isValidAuthority
     private static final String AUTHORITY_CHARS_REGEX = "\\p{Alnum}\\-\\."; // allows for IPV4 but not IPV6
-    // Allow for IPv4 mapped addresses: ::FFF:123.123.123.123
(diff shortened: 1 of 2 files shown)


garydgregory:
TY @sahvx655-wq , verified locally and merged 🚀

PR #407, opened 2026-06-22
Preserve fractional bound in BigIntegerValidator range checks

sahvx655-wq:
The Number-typed minValue and maxValue overrides convert both operands with toBigInteger before comparing, and toBigInteger truncates a non-integer bound towards zero. A fractional bound therefore loses its fractional part before the test, so the verdict is wrong: minValue(BigInteger 5, 5.9) returns true although 5 is below 5.9 (the bound is floored to 5), and maxValue(BigInteger -5, -5.9) returns true although -5 is above -5.9. I noticed this while checking these overloads against BigDecimalValidator, which already compares the same mixed Number bounds exactly, so the two validators disagree on identical inputs.

The fix compares both operands as BigDecimal so a non-integer bound keeps its value. Whole-number bounds and BigIntegers outside the long range are unaffected, and keeping the comparison inside the override mirrors BigDecimalValidator rather than pushing rounding onto callers. Left unfixed, any range check given a Double, Float or BigDecimal bound with a fractional part can admit values just outside the requested range.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
@@ -89,6 +89,10 @@ public class BigIntegerValidator extends AbstractNumberValidator {
         if (value instanceof Long) {
             return BigDecimal.valueOf(value.longValue());
         }
+        if (value instanceof Double) {
+            // No need to roundtrip with a string.
+            return BigDecimal.valueOf(((Double) value).doubleValue());
+        }
         return new BigDecimal(value.toString());
     }
 
--- a/src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
@@ -79,6 +79,19 @@ public class BigIntegerValidator extends AbstractNumberValidator {
         return VALIDATOR;
     }
 
+    private static BigDecimal toBigDecimal(final Number value) {
+        if (value instanceof BigDecimal) {
+            return (BigDecimal) value;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
```diff
@@ -79,6 +79,19 @@ public static BigIntegerValidator getInstance() {
...
     }
 
+    private static BigDecimal toBigDecimal(final Number value) {
+        if (value instanceof BigDecimal) {
+            return (BigDecimal) value;
+        }
+        if (value instanceof BigInteger) {
+            return new BigDecimal((BigInteger) value);
+        }
+        if (value instanceof Long) {
```
garydgregory, line 89: Hello @sahvx655-wq 
You missed the java.math.BigDecimal.valueOf(double) shortcut here.

garydgregory (changes requested):
@sahvx655-wq 
Thank you for the PR.
Please see my comment.

src/main/java/org/apache/commons/validator/routines/BigIntegerValidator.java
sahvx655-wq: Good catch. I've added a Double branch using BigDecimal.valueOf(((Double) value).doubleValue()) so it matches the toBigInteger helper just below, which already takes that path with the same "no need to roundtrip with a string" note. The fall-through stays on new BigDecimal(value.toString()) for the remaining types (Float keeps its own canonical short form that way rather than picking up the widened double representation).

Pushed in 3b48460 and the range tests still pass.

PR #408, opened 2026-06-22
compare week-year not calendar year in compareWeeks

sahvx655-wq:
compareWeeks reports 31 December and 1 January of the same calendar year as the same week. AbstractCalendarValidator.compare ordered the WEEK_OF_YEAR comparison on Calendar.YEAR and then the week number, but a week belongs to its own week-year, and at the turn of the year that week-year differs from the calendar year. Under the US locale 31 Dec 2018 has WEEK_OF_YEAR 1 and a week-year of 2019, so against 1 Jan 2018 (also WEEK_OF_YEAR 1) the year and the week number both match and two dates roughly 52 weeks apart compare equal. The mirror case fails the other way: 31 Dec 2018 and 1 Jan 2019 share one week yet compare as different weeks because their calendar years differ. WEEK_OF_MONTH resets the same way at the start of a month.

Following the review, this now orders both WEEK_OF_YEAR and WEEK_OF_MONTH by the distance in days first: dates a week or more apart are always in different weeks, and nearer dates share a week only when the week number also matches. That removes the year-boundary false match for WEEK_OF_YEAR and is correct by construction for WEEK_OF_MONTH, which keeps its previous results. It also drops the earlier getWeekYear() lookup and its unsupported-calendar fallback. Tests cover both fields across the year boundary and the full suite passes.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/AbstractCalendarValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/AbstractCalendarValidator.java
@@ -39,11 +39,8 @@ public abstract class AbstractCalendarValidator extends AbstractFormatValidator 
 
     private static final long serialVersionUID = -1410008585975827379L;
 
-    /** Number of days in a week, beyond which two dates cannot share a week. */
-    private static final int DAYS_PER_WEEK = 7;
-
-    /** Number of milliseconds in a day, used to convert an instant to a day count. */
-    private static final long MILLIS_PER_DAY = TimeUnit.DAYS.toMillis(1);
+    /** Number of milliseconds in a week, beyond which two instants cannot share a week. */
+    private static final long MILLIS_PER_WEEK = TimeUnit.DAYS.toMillis(7);
 
     /**
      * The date style to use for Locale validation.
@@ -426,9 +423,9 @@ public abstract class AbstractCalendarValidator extends AbstractFormatValidator 
      * Compares the week two calendars fall in, ordering by the actual week rather than by the
      * {@code WEEK_OF_YEAR} or {@code WEEK_OF_MONTH} number alone. Those numbers repeat across the
      * boundaries they reset on (for example 31 December may be week 1 of the following year, and
-     * the first week of a month can hold days carried over from the previous month), so the day
-     * distance is checked first: dates a week or more apart are always in different weeks, and
-     * nearer dates share a week only when the week number also matches.
+     * the first week of a month can hold days carried over from the previous month), so the gap
+     * between the two instants is checked first: dates a week or more apart are always in different
+     * weeks, and nearer dates share a week only when the week number also matches.
      *
      * @param value The Calendar value.
      * @param compare The {@link Calendar} to check the value against.
@@ -436,23 +433,10 @@ public abstract class AbstractCalendarValidator extends AbstractFormatValidator 
      * @return Zero if both calendars are in the same week, -1 or +1 otherwise.
      */
     private int compareWeek(final Calendar value, final Calendar compare, final int field) {
-        final long days = epochDay(value) - epochDay(compare);
-        if (Math.abs(days) >= DAYS_PER_WEEK || calculateCompareResult(value, compare, field) != 0) {
-            return Long.signum(days);
+        final long millis = value.getTimeInMillis() - compare.getTimeInMillis();
+        if (Math.abs(millis) >= MILLIS_PER_WEEK || calculateCompareResult(value, compare, field) != 0) {
+            return Long.signum(millis);
         }
         return 0;
-    }
-
-    /**
-     * Returns the number of whole days from the epoch to the calendar's local date, so two
-     * calendars can be compared by date regardless of the time of day.
-     *
-     * @param calendar The calendar to convert.
-     * @return the day count from the epoch in the calendar's own time zone.
-     */
-    private static long epochDay(final Calendar calendar) {
-        final long localMillis = calendar.getTimeInMillis()
-                + calendar.get(Calendar.ZONE_OFFSET) + calendar.get(Calendar.DST_OFFSET);
-        return Math.floorDiv(localMillis, MILLIS_PER_DAY);
     }
 }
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/validator/routines/CalendarValidatorTest.java
```diff
@@ -232,6 +232,27 @@ void testCompare() {
...
+        final Calendar jan01y2019 = createCalendar(TimeZones.GMT, 20190101, noon);
+
+        // both are calendar year 2018 and week-of-year 1, but lie ~52 weeks apart
+        assertEquals(1, calValidator.compareWeeks(dec31y2018, jan01y2018), "Dec 31 2018 is weeks after Jan 1 2018");
+        assertEquals(-1, calValidator.compareWeeks(jan01y2018, dec31y2018), "Jan 1 2018 is weeks before Dec 31 2018");
+
+        // 31 Dec 2018 and 1 Jan 2019 share week 1 of week-year 2019, so they are the same week
+        assertEquals(0, calValidator.compareWeeks(dec31y2018, jan01y2019), "Dec 31 2018 is the same week as Jan 1 2019");
+    }
+
```
sebbASF, line 280: I think there needs to be a test with a calendar that returns false for isWeekDateSupported().

sebbASF (changes requested):

src/main/java/org/apache/commons/validator/routines/AbstractCalendarValidator.java
```diff
@@ -416,4 +422,14 @@ protected Object parse(String value, final String pattern, final Locale locale,
...
+
+    /**
+     * Returns the week-year of a calendar, which a {@code WEEK_OF_YEAR} value is relative to.
+     *
+     * @param calendar The calendar to read the week-year from.
+     * @return the week-year, or the calendar year if week dates are not supported.
+     */
+    private int weekYear(final Calendar calendar) {
+        return calendar.isWeekDateSupported() ? calendar.getWeekYear() : calendar.get(Calendar.YEAR);
+    }
```
sebbASF: I’m not sure it makes sense to return the YEAR here. AFAICT this effectively converts a WEEK_OF_YEAR comparison into a YEAR comparison. It might make more sense to throw instead.

sebbASF (changes requested):

sebbASF:
It has just occurred to me - the same issue may occur with WEEK_OF_MONTH.
AFAICT the first WEEK_OF_MONTH in January could contain one or more days from December of the previous year.

Maybe an alternative approach would be to initially compare days since the epoch. If the difference between the dates is less than a week, then compare WEEK_OF values. If they are the same, then return 0, otherwise return the result of the epoch comparison.

src/test/java/org/apache/commons/validator/routines/CalendarValidatorTest.java
sahvx655-wq: The new version doesn't call isWeekDateSupported() or getWeekYear() any more, so that branch is gone. It keys off the day distance and the week number instead, which avoids the unsupported-calendar case entirely. I've added boundary tests for WEEK_OF_YEAR and WEEK_OF_MONTH in its place.

src/main/java/org/apache/commons/validator/routines/AbstractCalendarValidator.java
sahvx655-wq: Fair point, that YEAR fallback was effectively turning the comparison into a year comparison. I've dropped weekYear() altogether. compareWeek() now compares the day gap first and only consults the week number when the dates are under a week apart, so there's nothing left to fall back from.

sahvx655-wq:
Took the epoch route you suggested. compare() now sends both WEEK_OF_YEAR and WEEK_OF_MONTH through one helper that looks at the day gap first: a week or more apart is always a different week, and within a week they only count as the same week when the week number also matches. That removes the year-boundary false match on WEEK_OF_YEAR and drops the getWeekYear()/isWeekDateSupported() machinery the earlier version leaned on.

On WEEK_OF_MONTH: the old path compared MONTH before WEEK_OF_MONTH, so a December day and a January day never collided the way the year-boundary case did, which is why it wasn't visibly broken. Routing it through the same guard keeps every existing WEEK_OF_MONTH result and makes that invariant explicit rather than incidental. I added boundary tests for both fields and the full build (mvn clean verify with checkstyle/spotbugs/pmd, 1092 tests) is green.

sebbASF:
There’s another issue here - the WEEK_OF_* values depend on the Calendar settings, so only make sense if the calendars are the same according to .equals().

Does it make ever make sense to compare calendars in different time zones?

However, that can be handled separately.

src/main/java/org/apache/commons/validator/routines/AbstractCalendarValidator.java
```diff
@@ -416,4 +421,38 @@ protected Object parse(String value, final String pattern, final Locale locale,
...
+
+    /**
+     * Returns the number of whole days from the epoch to the calendar's local date, so two
+     * calendars can be compared by date regardless of the time of day.
+     *
+     * @param calendar The calendar to convert.
+     * @return the day count from the epoch in the calendar's own time zone.
+     */
+    private static long epochDay(final Calendar calendar) {
+        final long localMillis = calendar.getTimeInMillis()
```
sebbASF: getTimeInMillis returns the time as UTC milliseconds from the epoch, so uses the same base for both calendars.

Rather than calculating the day that this represents, it would be simpler to see if the times are more than a week of milliseconds apart.

sebbASF (changes requested):

src/main/java/org/apache/commons/validator/routines/AbstractCalendarValidator.java
sahvx655-wq: Agreed. getTimeInMillis() is already UTC-based for both calendars, so the local-day conversion was just noise. compareWeek() now subtracts the two getTimeInMillis() values directly, treats anything a week of millis or more apart as a different week, and inside that window defers to the week number. epochDay() and the day constants are gone.

Existing WEEK_OF results are unchanged, since the test calendars sit at noon in a single zone so the gaps come out as exact day multiples, and the full build is still green.

sahvx655-wq:
On the WEEK_OF_* values depending on calendar settings, and comparing across time zones: I agree those only really hold up when the calendars match, but since you noted it can be handled separately I've kept it out of this PR so the change stays scoped to the week-boundary fix. Happy to follow up on it on its own.

sebbASF:
Thanks!

PR #409, opened 2026-06-23
classify xn--90ae (бг) as a country-code TLD

sahvx655-wq:
DomainValidator keeps the IANA TLD lists as sorted arrays and decides a TLD's category from which array holds it. While going over the IDN country-code entries I noticed бг (punycode xn--90ae) sitting in GENERIC_TLDS, even though its Cyrillic neighbours срб and бел are in COUNTRY_CODE_TLDS. бг is the ccTLD delegated to Bulgaria for the two-letter code BG, so isValidCountryCodeTld("xn--90ae") and isValidCountryCodeTld("бг") wrongly return false while isValidGenericTld("xn--90ae") returns true.

Moving the entry into COUNTRY_CODE_TLDS, in its sorted position so the binary-search lookup still holds, makes the per-category checks agree with the IANA delegation. isValidTld and isValid already returned true through the category OR, so only the category-specific methods change behaviour. Left as is, callers that branch on TLD category (treating ccTLDs differently from gTLDs) mishandle Bulgarian domains. Added the matching assertions to DomainValidatorTest.testTopLevelDomains.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/DomainValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/DomainValidator.java
@@ -1370,7 +1370,6 @@ public class DomainValidator implements Serializable {
         "xn--80asehdb", // онлайн CORE Association
         "xn--80aswg", // сайт CORE Association
         "xn--8y0a063a", // 联通 China United Network Communications Corporation Limited
-        "xn--90ae", // бг Imena.BG Plc (NAMES.BG Plc)
         "xn--9dbq2a", // קום VeriSign Sarl
         "xn--9et52u", // 时尚 RISE VICTORY LIMITED
         "xn--9krt00a", // 微博 Sina Corporation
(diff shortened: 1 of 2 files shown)


sebbASF:
The lists are derived automatically from 
https://data.iana.org/TLD/tlds-alpha-by-domain.txt
and
https://www.iana.org/domains/root/db

It looks like xn--90ae is somehow not being classified correctly.

Any fixes to the table need to be done by correcting the logic, otherwise they may be reverted when the lists are next regenerated.

I'll have a look and see if I can find the problem.
Meanwhile, marking as draft because the PR is incomplete.

sebbASF:
Found the issue - the test did not check if the TLD was in the correct list.
If a TLD was originally added to the GENERIC list, changes were not detected.

sebbASF:
Thanks for the report!

PR #410, opened 2026-06-24
Compare exact values in DoubleValidator range checks

sahvx655-wq:
The `Number`-typed range overloads on `DoubleValidator` are inherited from `AbstractNumberValidator`, which tests `value.doubleValue()` against `bound.doubleValue()`. That narrows the bound to a `double` before comparing, so a `BigDecimal` or `BigInteger` bound carrying more significant digits than a `double` can hold is rounded onto the value first. `minValue(2^53, BigInteger 2^53 + 1)` returns `true` and `maxValue(2^53, 2^53 - 0.5)` returns `true`, both wrong: a value that sits outside the bound is reported as in range, and it then slips through any `isInRange` guard built on these methods.

The two overrides defer to the superclass `compareTo(Number, Number)`, so the bound's precision survives; this mirrors the `compareTo` comparisons already used in `BigIntegerValidator` and `BigDecimalValidator`. A non-finite operand keeps the old `doubleValue()` comparison so the documented infinity and NaN behaviour is unchanged. Placing the check in the callee means `isInRange(Number, Number, Number)` picks it up through virtual dispatch rather than each caller needing its own guard.

This was originally paired with the same change to `FloatValidator`, but I've dropped that half: at the magnitude where the bound-narrowing bug bites (around 2^53) a `float` spans roughly 10^9 per step, so the bound that triggers the bug always lands inside the value's round-trip interval, where `Float.toString` (which drives `compareTo`) differs between JDKs. That made the float regression test pass on JDK 21 but fail on JDK 8, and there's no portable way to pin it. `DoubleValidator` has no such window, so I've scoped this PR to it.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/DoubleValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/DoubleValidator.java
@@ -184,6 +184,42 @@ public class DoubleValidator extends AbstractNumberValidator {
     }
 
     /**
+     * Tests if the value is less than or equal to a maximum, comparing the exact values.
+     *
+     * <p>
+     * This overrides the {@link Number} overload inherited from the superclass, which narrows the bound to a {@code double} before comparing and so loses
+     * precision for a {@code BigDecimal} or {@code BigInteger} bound that carries more significant digits than a {@code double} can hold. A non-finite
+     * {@link Double} or {@link Float} operand keeps the {@code doubleValue()} comparison so the documented infinity behaviour is unchanged.
+     * </p>
+     *
+     * @param value The value validation is being performed on.
+     * @param max   The maximum value.
+     * @return {@code true} if the value is less than or equal to the maximum.
+     */
+    @Override
+    public boolean maxValue(final Number value, final Number max) {
+        return isFinite(value) && isFinite(max) ? compareTo(value, max) <= 0 : value.doubleValue() <= max.doubleValue();
+    }
+
+    /**
+     * Tests if the value is greater than or equal to a minimum, comparing the exact values.
+     *
+     * <p>
+     * This overrides the {@link Number} overload inherited from the superclass, which narrows the bound to a {@code double} before comparing and so loses
+     * precision for a {@code BigDecimal} or {@code BigInteger} bound that carries more significant digits than a {@code double} can hold. A non-finite
+     * {@link Double} or {@link Float} operand keeps the {@code doubleValue()} comparison so the documented infinity behaviour is unchanged.
+     * </p>
+     *
+     * @param value The value validation is being performed on.
+     * @param min   The minimum value.
+     * @return {@code true} if the value is greater than or equal to the minimum.
+     */
+    @Override
+    public boolean minValue(final Number value, final Number min) {
+        return isFinite(value) && isFinite(min) ? compareTo(value, min) >= 0 : value.doubleValue() >= min.doubleValue();
+    }
+
+    /**
      * Convert the parsed value to a {@code Double}.
      *
      * @param value The parsed {@code Number} object created.
--- a/src/test/java/org/apache/commons/validator/routines/DoubleValidatorTest.java
+++ b/src/test/java/org/apache/commons/validator/routines/DoubleValidatorTest.java
@@ -136,8 +136,8 @@ class DoubleValidatorTest extends AbstractNumberValidatorTest {
 
     /**
      * Test the {@link Number} range checks against a bound that carries more precision than a {@code double}.
-     * 2^53 is the largest integer with an exact {@code double} representation, so 2^53 + 1 cannot be narrowed
-     * onto the value: a value of 2^53 is below a minimum of 2^53 + 1 and above a maximum of 2^53 - 0.5.
+     * {@code 2^53} is the largest integer with an exact {@code double} representation, so {@code 2^53} + 1 cannot be narrowed
+     * onto the value: a value of {@code 2^53} is below a minimum of {@code 2^53 + 1} and above a maximum of {@code 2^53 - 0.5}.
      */
     @Test
     void testDoubleNumberRangeExactBound() {
--- a/src/test/java/org/apache/commons/validator/routines/DoubleValidatorTest.java
+++ b/src/test/java/org/apache/commons/validator/routines/DoubleValidatorTest.java
@@ -22,6 +22,8 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.text.DecimalFormatSymbols;
 import java.util.Locale;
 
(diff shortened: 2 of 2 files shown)


src/main/java/org/apache/commons/validator/routines/FloatValidator.java
```diff
@@ -68,6 +69,16 @@ public class FloatValidator extends AbstractNumberValidator {
 
     private static final FloatValidator VALIDATOR = new FloatValidator();
 
+    private static boolean isFinite(final Number value) {
```
garydgregory: Rebase on git master and reuse AbstractNumberValidator.isFinite(Number).

src/main/java/org/apache/commons/validator/routines/DoubleValidator.java
```diff
@@ -68,6 +69,16 @@ public class DoubleValidator extends AbstractNumberValidator {
 
     private static final DoubleValidator VALIDATOR = new DoubleValidator();
 
+    private static boolean isFinite(final Number value) {
```
garydgregory: Rebase on git master and reuse AbstractNumberValidator.isFinite(Number).

src/test/java/org/apache/commons/validator/routines/DoubleValidatorTest.java
```diff
@@ -132,6 +134,26 @@ void testDoubleRangeMinMaxNaN() {
...
+     * onto the value: a value of 2^53 is below a minimum of 2^53 + 1 and above a maximum of 2^53 - 0.5.
+     */
+    @Test
+    void testDoubleNumberRangeExactBound() {
+        final DoubleValidator validator = (DoubleValidator) strictValidator;
+        final long maxExactInt = 1L << 53; // 2^53
+        final Double value = Double.valueOf(maxExactInt);
+        final BigInteger above = BigInteger.valueOf(maxExactInt).add(BigInteger.ONE); // 2^53 + 1
+        final BigInteger below = BigInteger.valueOf(maxExactInt).subtract(BigInteger.ONE); // 2^53 - 1
+        final BigDecimal justBelow = new BigDecimal(BigInteger.valueOf(maxExactInt)).subtract(BigDecimal.valueOf(0.5)); // 2^53 - 0.5
```
garydgregory: Why is this not just `BigDecimal.valueOf(maxExactInt)` instead of `new BigDecimal(BigInteger.valueOf(maxExactInt))` but since this is Double validator instead of a "Big" validator is might be OK...

src/main/java/org/apache/commons/validator/routines/DoubleValidator.java
```diff
@@ -183,6 +194,42 @@ public boolean minValue(final Double value, final double min) {
...
+     * {@link Double} or {@link Float} operand keeps the {@code doubleValue()} comparison so the documented infinity behaviour is unchanged.
+     * </p>
+     *
+     * @param value The value validation is being performed on.
+     * @param max   The maximum value.
+     * @return {@code true} if the value is less than or equal to the maximum.
+     */
+    @Override
+    public boolean maxValue(final Number value, final Number max) {
+        return isFinite(value) && isFinite(max) ? BigDecimal.valueOf(value.doubleValue()).compareTo(toBigDecimal(max)) <= 0 : value.doubleValue() <= max.doubleValue();
```
garydgregory: Reuse the superclass' compareTo(Number, Number).

src/main/java/org/apache/commons/validator/routines/DoubleValidator.java
```diff
@@ -183,6 +194,42 @@ public boolean minValue(final Double value, final double min) {
...
+     * {@link Double} or {@link Float} operand keeps the {@code doubleValue()} comparison so the documented infinity behaviour is unchanged.
+     * </p>
+     *
+     * @param value The value validation is being performed on.
+     * @param min   The minimum value.
+     * @return {@code true} if the value is greater than or equal to the minimum.
+     */
+    @Override
+    public boolean minValue(final Number value, final Number min) {
+        return isFinite(value) && isFinite(min) ? BigDecimal.valueOf(value.doubleValue()).compareTo(toBigDecimal(min)) >= 0 : value.doubleValue() >= min.doubleValue();
```
garydgregory: Reuse the superclass' compareTo(Number, Number).

src/main/java/org/apache/commons/validator/routines/FloatValidator.java
```diff
@@ -183,6 +194,42 @@ public boolean minValue(final Float value, final float min) {
...
+     * {@link Double} or {@link Float} operand keeps the {@code doubleValue()} comparison so the documented infinity behaviour is unchanged.
+     * </p>
+     *
+     * @param value The value validation is being performed on.
+     * @param max   The maximum value.
+     * @return {@code true} if the value is less than or equal to the maximum.
+     */
+    @Override
+    public boolean maxValue(final Number value, final Number max) {
+        return isFinite(value) && isFinite(max) ? BigDecimal.valueOf(value.doubleValue()).compareTo(toBigDecimal(max)) <= 0 : value.doubleValue() <= max.doubleValue();
```
garydgregory: It looks like `BigDecimal.valueOf(value.doubleValue())` looses precision.

garydgregory (changes requested):
Hello @sahvx655-wq 
Please rebase on git master and see my comments.

src/main/java/org/apache/commons/validator/routines/DoubleValidator.java
sahvx655-wq: Done. Rebased on master and dropped the local helper in both DoubleValidator and FloatValidator so they use the inherited isFinite(Number).

src/main/java/org/apache/commons/validator/routines/DoubleValidator.java
sahvx655-wq: Switched over. Both minValue and maxValue in both validators now call the superclass compareTo(value, bound) instead of building the BigDecimal by hand.

src/main/java/org/apache/commons/validator/routines/FloatValidator.java
sahvx655-wq: Right, that was the wrong basis. Going through compareTo(value, max) runs the value through toBigDecimal, which for a Float uses Float.toString, so the float is compared at its own precision (9.007199E15) rather than the extra digits doubleValue() invents when it widens the float. I adjusted the FloatValidatorTest expectations to that value.

src/test/java/org/apache/commons/validator/routines/DoubleValidatorTest.java
sahvx655-wq: Simplified to BigDecimal.valueOf(maxExactInt) in the Double test. Same value, less noise.

sahvx655-wq:
Rebased on master and reused the pulled-up helpers throughout: both validators drop the local isFinite copy for the inherited isFinite(Number), and minValue/maxValue now defer to the superclass compareTo(Number, Number) rather than assembling the BigDecimal locally. The test bound is BigDecimal.valueOf(...) now too.

On the float precision point: you were right that BigDecimal.valueOf(value.doubleValue()) was wrong. compareTo runs the value through toBigDecimal, and for a Float that uses Float.toString, so the comparison sits at the float's own decimal value (9.007199E15) instead of the trailing digits doubleValue() fabricates when it widens the float. That moved the FloatValidatorTest expectations onto 9.007199E15, which is the honest bound for a float at that magnitude. mvn clean verify is green.

sahvx655-wq:
Sorry this sat red for a week. Chased down the CI failure: only the JDK 8 job actually failed, the rest of the matrix was cancelled by fail-fast, so it looked worse than it was. The culprit was my own testFloatNumberRangeExactBound. A float at 2^53 has a round-trip interval about a float ULP wide (~10^9 there), and compareTo runs the value through toBigDecimal, which for a Float goes via Float.toString. JDK 19+ emits the shortest form (9.007199E15) while JDK 8's older algorithm emits the exact 2^53, and my bound sat between the two, so the assertion flipped by JDK.

The deeper problem is that any bound precise enough to trigger the double-narrowing bug has to fall within one double ULP of the value, which at that magnitude is always inside the float's round-trip interval, so there's no portable way to pin the float override in a test. I've dropped FloatValidator and scoped the PR to Double, where doubleValue() is exact and the test is stable (it already passed on JDK 8). The Double overrides defer to the superclass compareTo(Number, Number) and the inherited isFinite(Number) as you asked, and the test bound is BigDecimal.valueOf(...). mvn clean verify is green locally.

src/main/java/org/apache/commons/validator/routines/FloatValidator.java
sahvx655-wq: Dropped the FloatValidator change from this PR, so this thread is moot. Reasoning in the PR comment: the float override can't be pinned by a JDK-portable test at 2^53.

src/main/java/org/apache/commons/validator/routines/FloatValidator.java
sahvx655-wq: Removed the FloatValidator override entirely, so this is no longer in play. The precision point is exactly why it can't be tested portably at that magnitude; details in the PR comment.

PR #411, opened 2026-06-24
trim input in IBANValidator.validate

sahvx655-wq:
Noticed while validating IBANs read line by line out of a file: a value with a trailing newline came back INVALID_LENGTH and one with a leading space UNKNOWN_COUNTRY, although both validate once the surrounding whitespace is stripped.

1. validate(String) is the only validation method here and isValid delegates to it, yet it never normalises its argument, unlike CodeValidator.validate and the AbstractNumberValidator/AbstractCalendarValidator parse paths which all trim first.
2. the raw string is handed to getValidator, which reads the first two characters as the country code, so a leading space selects the wrong code and any surrounding whitespace also trips the fixed-length check, leaving the returned IBANValidatorStatus misleading rather than VALID.

Trimmed once at the top of validate and ran the country lookup, length, pattern and check digit against that value; the null path is unchanged as getValidator already tolerates null.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/IBANValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/IBANValidator.java
@@ -417,16 +417,17 @@ public class IBANValidator {
      * @since 1.10.0
      */
     public IBANValidatorStatus validate(final String code) {
-        final Validator formatValidator = getValidator(code);
+        final String iban = code == null ? null : code.trim();
+        final Validator formatValidator = getValidator(iban);
         if (formatValidator == null) {
             return IBANValidatorStatus.UNKNOWN_COUNTRY;
         }
-        if (code.length() != formatValidator.ibanLength) {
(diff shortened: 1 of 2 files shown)


sebbASF:
Sorry, but I don't think this is a good idea.

The purpose of validation is to check the data that is passed to it.

If a caller passes an IBAN with leading or trailing spaces, and the result is success, then the caller should be able to expect that the value can be used as is when passing the data to another method which expects an actual IBAN without any spaces.

Any trimming needs to be done by the caller.

sahvx655-wq:
Fair enough, that's a reasonable line to draw. The contract argument holds: if validate() trims and returns VALID, a caller could quite reasonably treat the string it handed in as a clean IBAN and pass it on with the whitespace still attached, which is worse than a clear rejection. Leaving normalisation to the caller keeps the status honest about exactly what was given.

I'll leave this closed. Cheers for taking a look.

PR #412, opened 2026-06-25
range-check the port for IPv6 hosts in isValidAuthority

sahvx655-wq:
**IPv6 host port skips the range check in isValidAuthority**

The port group in `AUTHORITY_PATTERN` is captured the same way for a bracketed IPv6 host as for a hostname or IPv4 host, but the 0 to 65535 range check only ran in the hostname/IPv4 branch of `isValidAuthority`. That left a gap: `http://[::1]:99999/` and `http://[::1]:65536/` validated as good while the equivalent `http://127.0.0.1:99999/` and `http://example.com:99999/` were correctly rejected, so an out-of-range port slipped through purely on the strength of the host being IPv6.

Moved the existing port parse and range check out of that branch so both host forms are checked the same way. Valid ports such as `[::1]:65535` and `[::1]:80`, and the no-port case, are unaffected. Added a regression test that fails without the change.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
@@ -443,16 +443,19 @@ public class UrlValidator implements Serializable {
                     return false;
                 }
             }
-            final String port = authorityMatcher.group(PARSE_AUTHORITY_PORT);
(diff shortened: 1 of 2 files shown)


sebbASF:
Thanks!

PR #413, opened 2026-06-25
Fix swapped constant maps in Form.process for extended forms

sahvx655-wq:
I was reading through the form-extension path and noticed `Form.process` calls itself with the first two arguments in a different order to the method's own signature and to the other two call sites, which is what put me onto this.

1. `Form.process` recurses into the inherited form at Form.java:224 as `parent.process(constants, globalConstants, forms)`, so the global map and the form-set map are swapped relative to the declared `process(globalConstants, constants, forms)`.
2. `Field.process` applies the form-set constants first and the global constants second, so for a shared constant name the form-set value is meant to win; because of the swap a parent form reached through `extends` resolves that constant to the global value instead, the opposite precedence to a non-extended form and to the correct calls at Form.java:240 and FormSet.java:312.
3. Restored the declared argument order on the recursive call, and added a regression test in `ExtensionTest` that extends a base form sharing a constant name with a global one and asserts the form-set value wins; it fails before this change (`expected: <formset> but was: <global>`) and passes after.

Left unfixed, any form that uses `extends` together with a form-set constant overriding a like-named global constant silently picks up the wrong value, and which value wins also depends on the form iteration order, so it is not deterministic across runs.

### DIFF
--- a/src/main/java/org/apache/commons/validator/Form.java
+++ b/src/main/java/org/apache/commons/validator/Form.java
@@ -221,7 +221,7 @@ public class Form implements Serializable {
             if (parent != null) {
                 if (!parent.isProcessed()) {
                     // we want to go all the way up the tree
-                    parent.process(constants, globalConstants, forms);
+                    parent.process(globalConstants, constants, forms);
                 }
                 for (final Field f : parent.getFields()) {
                     // we want to be able to override any fields we like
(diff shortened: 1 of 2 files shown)


garydgregory:
Good catch @sahvx655-wq , merged 🚀

PR #415, opened 2026-07-03
Reject vowels in SedolCheckDigit

sahvx655-wq:
`SedolCheckDigit.toInt` already rejects out-of-range and non-alphanumeric characters, but it accepts vowels. The SEDOL alphabet deliberately omits A, E, I, O and U (as noted in the Wikipedia reference in the class Javadoc), so no genuine SEDOL contains one. Whilst working through these routines I generated a correct modulus-10 check digit for bodies containing vowels and found `SEDOL_CHECK_DIGIT.isValid` returning true for codes such as `B0AKT02` and `AAAAAA0`. There is no `SedolValidator` wrapping this routine, so `SedolCheckDigit.isValid` is the only public SEDOL entry point and the over-acceptance is user visible: a string that can never be a real SEDOL passes validation purely because its check digit happens to line up.

The fix rejects vowels in `toInt`, next to the existing alphanumeric-range check, which is where the other per-character SEDOL rules already live, so the whole character set stays enforced in one place rather than being pushed onto callers. The added test asserts vowel-bearing codes with valid check digits are rejected while the existing consonant fixtures still validate; it fails before the change and passes after.

- [x] Run a successful build using the default Maven goal with `mvn`.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/checkdigit/SedolCheckDigit.java
+++ b/src/main/java/org/apache/commons/validator/routines/checkdigit/SedolCheckDigit.java
@@ -89,7 +89,15 @@ public final class SedolCheckDigit extends ModulusCheckDigit {
         if (charValue > charValueMax || !isAsciiAlphaNum(character)) {
             throw new CheckDigitException("Invalid Character[%d,%d] = '%d' out of range 0 to %d", leftPos, rightPos, charValue, charValueMax);
         }
+        // The SEDOL alphabet omits the vowels A, E, I, O and U, so a code containing one is not a SEDOL.
+        if (isVowel(character)) {
+            throw new CheckDigitException("Invalid Character[%d,%d] = '%c' - vowels are not used in a SEDOL", leftPos, rightPos, character);
+        }
         return charValue;
+    }
+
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hello @sahvx655-wq 

Thank you for the PR.

- Use a parameterized test instead of a loop
- Add test fixtures that check each vowel indidivually.

TY!

sahvx655-wq:
Makes sense. Switched testVowelsRejected to a @ParameterizedTest driven by @ValueSource and dropped the loop. Added a fixture per vowel that isolates it: A000000, E000006, I000002, O000006 and U000000, each a single vowel in the body with a correct modulus 10 check digit, so they only fail once the vowel exclusion is applied rather than on a check-digit mismatch. The earlier multi-vowel codes (B0AKT02, 0EIOU02, BAEIOU7, AAAAAA0) are kept as further fixtures.

Each row now runs as its own invocation, so a regression on one particular vowel points straight at that vowel instead of failing the whole loop. Full build stays green.

garydgregory:
@sahvx655-wq 
You forgot to update the Javadoc, I'll fix that in git master.

PR #416, opened 2026-07-04
Compare exact bound in AbstractNumberValidator range checks

sahvx655-wq:
The DoubleValidator range fix in #410 left the same narrowing in the shared parent. `AbstractNumberValidator.minValue(Number, Number)` and `maxValue(Number, Number)` fall back to `value.longValue()` against `min.longValue()` on the integer branch, and `ByteValidator`, `ShortValidator`, `IntegerValidator` and `LongValidator` inherit that branch without overriding it, so a bound that does not fit in a long is mangled before the comparison ever happens. `LongValidator.maxValue(Long.MAX_VALUE, 2^63 held as a BigInteger)` narrows the bound to `Long.MIN_VALUE` and reports the in-range value as over the maximum, while a fractional bound such as `minValue(5, 5.5)` is floored to 5 and wrongly passes. The visible behaviour is a validator quietly comparing against the wrong bound whenever a small integer value meets a `BigInteger`, `BigDecimal` or fractional limit, which is easy to miss because no error is raised.

The `BigInteger`, `BigDecimal` and `Double` overloads already compare the operands exactly and keep the `doubleValue()` fallback for a non-finite bound, so I put the same treatment on the parent's integer branch rather than repeat an override in all four subclasses. Sitting the fix in the shared base keeps the `NaN` and infinity handling identical across the hierarchy and avoids a fifth copy of the comparison. Each of the four affected validator tests gets a regression that fails on the old narrowing branch and passes now.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/AbstractNumberValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/AbstractNumberValidator.java
@@ -272,6 +272,12 @@ public abstract class AbstractNumberValidator extends AbstractFormatValidator {
     /**
      * Tests if the value is less than or equal to a maximum.
      *
+     * <p>
+     * For an integer validator the bound is compared exactly rather than narrowed to a {@code long}, so a {@code BigInteger} or {@code BigDecimal} bound outside
+     * the long range keeps its magnitude and a fractional bound keeps its fractional part. A non-finite {@link Double} or {@link Float} operand keeps the
+     * {@code doubleValue()} comparison, since {@code BigDecimal} cannot represent {@code NaN} or an infinity.
+     * </p>
+     *
      * @param value The value validation is being performed on.
      * @param max   The maximum value.
      * @return {@code true} if the value is less than or equal to the maximum.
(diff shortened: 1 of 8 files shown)


garydgregory:
Hello @sahvx655-wq 
The PR changes an abstract main class. Make sure ALL concrete subclasses of the abstract class have tests that check the changes in main.
TY!

sahvx655-wq:
Makes sense. I'd only added the regression to the four integer validators that route straight through the changed non-fraction branch (Byte/Short/Integer/Long), but you're right that every concrete subclass should exercise the inherited overloads. I've now added the same testNumberRangeExactBound to FloatValidatorTest, CurrencyValidatorTest and PercentValidatorTest. Float hits the unchanged fraction branch and Currency/Percent inherit BigDecimalValidator's exact-compare override, so those assert the behaviour stays correct through each concrete class rather than failing on the old tree.

Double, BigDecimal and BigInteger already carry equivalent coverage from the earlier range fixes (testDoubleNumberRangeExactBound, testNumberRangeBeyondDoublePrecision, and testNumberRangeOutsideLongRange/testNumberRangeFractionalBound/testNumberRangeNonFiniteBound), so the whole hierarchy is now covered. Full mvn build is green.

PR #419, opened 2026-07-08
Handle Float operand in AbstractNumberValidator.toBigDecimal

sahvx655-wq:
**Float bound read at the wrong magnitude in the number range checks**

While extending the exact-bound range checks to a `Float` limit I found `BigIntegerValidator.getInstance().maxValue(new BigInteger("100000000000000000001"), 1e20f)` returning false, even though the float `1e20f` actually holds 100000002004087734272 and the value sits below it. The `Number` range overloads convert both operands through `AbstractNumberValidator.toBigDecimal` before comparing, and that helper branches on `Long`, `Double`, `BigDecimal` and `BigInteger` but has no `Float` case, so a `Float` drops to the final `new BigDecimal(value.toString())`. `Float.toString` emits only the digits needed to round-trip the value, so the bound is rebuilt from "1.0E20" and compared as 10^20 rather than its real magnitude, and anything between the two is mis-ranged.

Root cause is the missing branch, so I widened a `Float` through `doubleValue()` and used `BigDecimal.valueOf`, mirroring the neighbouring `Double` branch, and the operand is now compared at the value it truly holds. If left as is a value near a float limit is silently admitted or rejected across `BigIntegerValidator`, `BigDecimalValidator` and `DoubleValidator`, and `toBigInteger` carries the same skew since it routes through `toBigDecimal`. The regression test in `BigIntegerValidatorTest` derives its fixture from the float itself so it does not lean on a hand-written constant.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/AbstractNumberValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/AbstractNumberValidator.java
@@ -94,6 +94,9 @@ public abstract class AbstractNumberValidator extends AbstractFormatValidator {
         if (value instanceof Double) {
             return BigDecimal.valueOf(((Double) value).doubleValue());
         }
+        if (value instanceof Float) {
+            return BigDecimal.valueOf(((Float) value).doubleValue());
+        }
         if (value instanceof BigDecimal) {
             return (BigDecimal) value;
         }
--- a/src/test/java/org/apache/commons/validator/routines/BigIntegerValidatorTest.java
+++ b/src/test/java/org/apache/commons/validator/routines/BigIntegerValidatorTest.java
@@ -221,6 +221,27 @@ class BigIntegerValidatorTest extends AbstractNumberValidatorTest {
     }
 
     /**
+     * A {@link Float} bound must be compared at its real magnitude. The bound is converted through {@link AbstractNumberValidator#toBigDecimal}, which had no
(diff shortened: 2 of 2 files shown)


hiSandog:
Handling `Float` separately is important because its decimal round-trip string is not necessarily the exact widened value used in numeric comparison. In addition to the positive upper-bound regression, a mirrored negative-bound case near `-1e20f` would help catch sign-sensitive behavior in both `minValue` and `maxValue`. A test using `Float.NaN` or infinity would also clarify whether these non-finite operands are intentionally rejected or delegated to the existing exception path.

sahvx655-wq:
Both added. The negative mirror is worth having since the skew flips sign: -1e20f prints closer to zero than the value it actually holds, so on the unpatched path a value sitting between the two magnitudes was wrongly admitted by maxValue and wrongly failed minValue, the opposite direction to the positive case. The new block in testNumberRangeFloatBound derives its fixture from the float itself, same as the positive case.

On the non-finite side I traced the call path first: a Float NaN or infinity never reaches toBigDecimal, because the isFinite guard in the range overloads already branches on Float as well as Double, so those operands are delegated to the plain double comparison rather than rejected with an exception. In practice a NaN bound is never satisfied and an infinity behaves as an open bound; I've extended testNumberRangeNonFiniteBound to assert that for Float explicitly so the delegation is pinned down rather than implied. Full suite still green, 1113 tests.

PR #423, opened 2026-07-15
Reject fractional input in GenericTypeValidator integer helpers

sahvx655-wq:
Following the formatLong range-check fix, formatByte, formatInt and formatShort (the String, Locale overloads) still range-check the parsed number with num.doubleValue() and then narrow it with byteValue()/intValue()/shortValue(). The integer-only NumberFormat only stops parsing at the decimal separator, so a fractional value written with a negative exponent and no decimal point is consumed whole and handed back as a Double: under Locale.US "15E-1" (1.5) satisfies the double bounds and byteValue() floors it, so formatByte returns 1 and formatInt("5E-1") returns 0 instead of null. That earlier fix observed the sibling bounds are exact as doubles, which rules out the long magnitude gap but not this fractional one, so the three helpers report an integer the caller never entered.

Guard each of the three on num instanceof Long before the per-type range check, which NumberFormat returns only for a non-fractional value in long range, matching formatLong and routines.Byte/Short/IntegerValidator. Keeping the decision inside the conversion helper means a value that was never integral comes back as null rather than a silently truncated one. Added GenericTypeValidatorTest#testIntegerLocaleFractional, which fails on the current code and passes with the change.

### DIFF
--- a/src/main/java/org/apache/commons/validator/GenericTypeValidator.java
+++ b/src/main/java/org/apache/commons/validator/GenericTypeValidator.java
@@ -66,9 +66,15 @@ public class GenericTypeValidator implements Serializable {
         if (value != null) {
             final ParsePosition pos = new ParsePosition(0);
             final Number num = getIntegerOnlyFormat(locale).parse(value, pos);
-            // If there was no error and we used the whole string
-            if (pos.getErrorIndex() == -1 && pos.getIndex() == value.length() && num.doubleValue() >= Byte.MIN_VALUE && num.doubleValue() <= Byte.MAX_VALUE) {
-                result = Byte.valueOf(num.byteValue());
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hello @sahvx655-wq 
Please split the new test into one method per API tested, `formatByte`, `formatShort`, `formatInt`. TY!

garydgregory:
@sahvx655-wq ping 🔔

sahvx655-wq:
Sorry for the slow reply, missed the notification. Done: split the test into testByteLocaleFractional, testShortLocaleFractional and testIntLocaleFractional, one method per API with the same assertions as before. Test class and checkstyle both pass locally.

PR #426, opened 2026-07-17
Reject over-length codes in ISSNCheckDigit

sahvx655-wq:
Noticed while exercising the check digit routines directly rather than through their wrapper validators.

1. `ISSNCheckDigit.ISSN_CHECK_DIGIT.isValid` is a public entry point and `ModulusCheckDigit` never checks the code length, so an over-length string reaches the modulus test unchecked.
2. `weightedValue` weights each position by `9 - leftPos`, so a ninth character lands on weight zero and contributes nothing; appending any character to a valid numeric ISSN (`031784710`..`031784719`, `0317847100`) still returns true, though only the eight-character `03178471` is a real ISSN.
3. Guarded `isValid` to require the fixed eight-character length before the check digit test, the same shape as the existing `IBANCheckDigit` length guard; the `ISSNValidator`/`CodeValidator` path already enforces length, so valid input is unaffected.

Regression test `testOverLengthRejected` added, which fails on the current code and passes with the guard.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/checkdigit/ISSNCheckDigit.java
+++ b/src/main/java/org/apache/commons/validator/routines/checkdigit/ISSNCheckDigit.java
@@ -52,6 +52,9 @@ public final class ISSNCheckDigit extends ModulusCheckDigit {
 
     private static final long serialVersionUID = 1L;
 
+    /** An ISSN is exactly eight characters: seven digits plus the check digit. */
+    private static final int ISSN_LEN = 8;
+
     /**
      * Singleton ISSN Check Digit instance.
      */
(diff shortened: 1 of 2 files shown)


garydgregory:
@sahvx655-wq 
See my comment https://github.com/apache/commons-validator/pull/423#pullrequestreview-4704982785

garydgregory:
Thank you @sahvx655-wq , merged 🚀 Please check for other length issues in this package.

sahvx655-wq:
Will do. From a quick look the mirror form exists in the rightPos-weighted routines: ISBN10CheckDigit and ABANumberCheckDigit have no length check either, and there a prepended zero is the character that lands on a no-op weight, so a leading-zero-padded code keeps the same sum. I'll work through the package and put up fixes for the ones reachable outside their wrapper validators. Thanks for merging this.

PR #428, opened 2026-07-21
Require the IPv6: tag for bracketed IPv6 email literals

sahvx655-wq:

`EmailValidator.isValidDomain` hands the contents of a bracketed host to `InetAddressValidator.isValid`, which accepts an IPv4 or an IPv6 address without distinguishing them. RFC 5321 section 4.1.3 tags an IPv6 address literal with `IPv6:` and leaves an IPv4 literal untagged, so the current code is wrong at both ends: it accepts a bare `user@[::1]`, which no conformant parser should, and rejects the correctly tagged `user@[IPv6:::1]` because the `IPv6:` prefix reaches `InetAddressValidator` as part of the address and the parse fails. I came across it while checking why a valid IPv6 literal was being turned away when the malformed bare form validated.

The fix matches the `IPv6:` tag on the bracket contents, case insensitively since it is ABNF-literal text, and validates the remainder as IPv6; an untagged literal is validated as IPv4 only. Keeping the split in `isValidDomain` leaves the IPv4-versus-IPv6 rule in the one place that parses the bracketed host, and the deprecated `validator.EmailValidator` inherits it because its `isValid` delegates here. Left unfixed, an allow-list built on this treats a bare loopback or link-local literal that no conformant MTA routes as valid while refusing the RFC form.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/EmailValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/EmailValidator.java
@@ -43,6 +43,8 @@ public class EmailValidator implements Serializable {
 
     private static final String EMAIL_REGEX = "^(.+)@(\\S+)$";
     private static final String IP_DOMAIN_REGEX = "^\\[(.*)\\]$";
+    // RFC 5321 section 4.1.3: an IPv6 address literal is prefixed with this tag, an IPv4 literal is not.
+    private static final String IPV6_TAG = "IPv6:";
     private static final String USER_REGEX = "^" + WORD + "(\\." + WORD + ")*$";
 
     private static final Pattern EMAIL_PATTERN = Pattern.compile(EMAIL_REGEX);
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
This is now a mix of RegEx and string matching. Shouldn't it be possible to update the regex and keep it at that?

sahvx655-wq:
Fair point, the regionMatches/substring pair was doing work the pattern could do itself. I've moved the tag into IP_DOMAIN_REGEX as an optional capture group, `^\[((?i)IPv6:)?(.*)\]$`, so the match itself records whether the tag was present and isValidDomain just picks isValidInet6Address or isValidInet4Address based on group(1), with group(2) as the address. The (?i) is scoped to the tag group so the address part is unaffected, and the IPV6_TAG constant and string handling are gone.

Behaviour is unchanged from the previous revision: tagged literals validate as IPv6 in any case, untagged ones as IPv4 only. The new test still passes and the full default mvn goal (rat, japicmp, checkstyle, spotbugs, pmd, javadoc) is BUILD SUCCESS.

PR #430, opened 2026-07-28
Drop the symbol separator in the currency and percent re-parse

sahvx655-wq:
Came at this from the other end: the German and French number validators were rejecting amounts the British one accepted, and the parse trace showed both the initial attempt and the lenient retry failing at the same index, which puts the fault in the fallback pattern rather than in the input.

1. `CurrencyValidator.parse` and `PercentValidator.parse` rebuild the pattern with the symbol removed and re-parse, but they delete only the symbol character itself.
2. `NumberFormat.getCurrencyInstance(Locale.GERMANY).toPattern()` is `#,##0.00 ¤` and `getPercentInstance(Locale.FRANCE).toPattern()` is `#,##0 %`, so what survives the rebuild is a mandatory non-breaking-space suffix.
3. The retry then fails exactly where the first parse did, so `validate("1.234,56", Locale.GERMANY)` and `validate("12", Locale.FRANCE)` come back null, while the prefix-symbol locales such as en-GB are fine.
4. The inverse is the worse half: a bare non-breaking space satisfies that leftover suffix on its own, so `validate("1.234,56 ", Locale.GERMANY)` yields 1234.56 and `validate("12 ", Locale.FRANCE)` yields 0.12 with no symbol present at all.

Both methods now drop the space adjacent to the symbol along with it. Left alone, the behaviour cuts both ways for every suffix-symbol locale (de, fr, it and the rest): correctly formatted amounts are silently refused, and a stray separator passes as a substitute for the currency or percent sign, which is a poor property in something callers lean on to reject malformed money. The two regression tests build an explicit suffix pattern instead of reading the JDK's locale data, so they will not drift with CLDR updates; both fail on master and pass with the change.

### DIFF
--- a/src/test/java/org/apache/commons/validator/routines/CurrencyValidatorTest.java
+++ b/src/test/java/org/apache/commons/validator/routines/CurrencyValidatorTest.java
@@ -20,15 +20,20 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
 import java.text.NumberFormat;
 import java.util.Locale;
+import java.util.stream.Stream;
 
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.junitpioneer.jupiter.DefaultLocale;
 
 /**
@@ -43,6 +48,15 @@ class CurrencyValidatorTest {
 
     private String usDollar;
     private String ukPound;
+
+    /**
+     * Locales whose currency format suffixes the symbol behind a space separator, covering different concrete symbols (&euro; and kr).
+     *
+     * @return the locales to test.
+     */
+    static Stream<Locale> suffixSymbolLocales() {
+        return Stream.of(Locale.GERMANY, Locale.FRANCE, Locale.forLanguageTag("sv-SE"));
+    }
 
     @BeforeEach
     protected void setUp() {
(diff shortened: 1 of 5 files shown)


src/test/java/org/apache/commons/validator/routines/PercentValidatorTest.java
```diff
@@ -100,6 +105,21 @@ void testNumberRangeExactBound() {
         assertFalse(instance.minValue(new BigDecimal("5"), new BigDecimal("5.5")));
     }
 
+    /**
+     * Test percentage values with a pattern that suffixes the symbol, which locales such as fr-FR separate from the number with a non-breaking space. The
```
garydgregory: The comment make a claim about FR and the code about US. Add a test for each, or parameterize the current test for both, better yet, for all known currency symbols.

src/test/java/org/apache/commons/validator/routines/CurrencyValidatorTest.java
```diff
@@ -176,6 +179,21 @@ void testPattern() {
         assertFalse(validator.isValid(ukPound + "1,234.567", pattern, Locale.US), "invalid symbol");
     }
 
+    /**
+     * Test currency values with a pattern that suffixes the symbol, which locales such as de-DE separate from the number with a non-breaking space. The symbol
```
garydgregory: The comment make a claim about DE and the code about US. Add a test for each, or parameterize the current test for both, better yet, for all known currency symbols.

garydgregory (changes requested):
@sahvx655-wq 
Thank you for the PR.
Please review my comments as well as the copilot comments.

src/test/java/org/apache/commons/validator/routines/CurrencyValidatorTest.java
sahvx655-wq: Fair point, the javadoc claimed de-DE while the assertions only ever ran against en-US symbols. There's now a parameterised testSuffixSymbolLocale covering de-DE, fr-FR and sv-SE against the JVM's own currency format, with the pattern, separator and inputs all derived from the locale data at run time; that matters because the separator has drifted between CLDR releases (fr moved from U+00A0 towards U+202F), so hardcoding any one character would either rot or fail on older JVMs. On a JVM whose data doesn't space-separate a suffix symbol the case skips via an assumption instead of asserting stale data, and the explicit-pattern US test stays as the pinned variant.

On covering all known currency symbols: at pattern level there is only the one placeholder, U+00A4, and the concrete sign is substituted per locale from DecimalFormatSymbols, so varying the locale is what varies the symbol. The new set exercises the euro and the Swedish krona alongside the dollar and pound already in this file. Each locale case fails on master with 'no symbol ==> expected: <1234.56> but was: <null>' and passes with the fix.

src/test/java/org/apache/commons/validator/routines/PercentValidatorTest.java
sahvx655-wq: Same treatment as the currency test: a parameterised testSuffixSymbolLocale now runs fr-FR and de-DE against NumberFormat.getPercentInstance with the pattern and separator read from the locale data at run time, and the explicit-pattern test keeps the pinned en-US expectations. The assumption guard means a JVM whose percent data has no space-separated suffix (the older legacy-data JREs) skips the case rather than failing on data the fix can't influence.

Both locale cases fail on master, 'no symbol: fr_FR ==> expected: <0.12> but was: <null>', and pass with the change, so the claim in the comment is now backed by an assertion rather than prose.

sahvx655-wq:
Pushed the review changes. The copilot points are folded in as well: the two private isNextToSymbol copies are gone, both validators now share a single package-private BigDecimalValidator.removeSymbol(pattern, symbol) with the adjacency checks explicitly parenthesised, and the test locals that shadowed the fixture field are renamed to instance. That also removes the pre-existing duplication of the strip loop itself, so the currency and percent paths can no longer drift apart.

Full default-goal build with -Ddoclint=all is green: 1154 tests, 0 failures, checkstyle, pmd, spotbugs and javadoc all clean. I also re-ran the new tests against master's validators to confirm every parameterised locale case fails without the fix before it passes with it.

sahvx655-wq:
Extended this per the 'better yet, for all known currency symbols' remark: the explicit-pattern tests are no longer pinned to en-US, they're parameterised over Locale.getAvailableLocales() with the inputs formatted through each locale's own DecimalFormatSymbols. That exercises the strip for every currency and percent symbol the JVM ships, including the multi-character symbols and the locales with non-Latin digits, without the expectations depending on any one locale's data.

Ran the sweep against master's validators first to make sure it still bites: all 2138 locale cases fail there with 'no symbol ==> expected: <1234.56> but was: <null>' and pass with the fix. Full default-goal build with -Ddoclint=all is green afterwards: 3290 tests, checkstyle, pmd and spotbugs all clean.

garydgregory:
@sahvx655-wq 
Run tests locally _before_ you push please. Build is broken.

sahvx655-wq:
Sorry about that, the suite was green here on 21 but I hadn't run it against the Java 8 job's locale data, which is where it broke. The CI trace shows the 'separator without symbol' assertion failing for de_DE, fr_FR and sv_SE plus the fr_FR percent case: the legacy pre-CLDR data separates the suffix symbol with a plain U+0020 space rather than a non-breaking one, and AbstractNumberValidator trims the input before parsing, so on Java 8 the bare separator degenerates to the valid no-symbol form and parses to 1234.56 instead of being rejected. The explicit-pattern sweep pins a non-breaking space, so it was unaffected.

The locale-data test now asserts to the trim contract: a separator above U+0020 with no symbol must be rejected, while a plain-space one is trimmed away and must parse as the no-symbol form. I reproduced the Java 8 failure locally with -Djava.locale.providers=COMPAT before changing anything, and after the fix both providers are green; the full default goal with -Ddoclint=all also passes, 3290 tests, checkstyle, pmd and spotbugs all clean.

PR #433, opened 2026-08-11
UrlValidator compare the file scheme case-insensitively in isValid

sahvx655-wq:
Chasing why an all upper case `FILE:` URL validated differently from its lower case twin led back to `isValid`. `isValidScheme` matches the scheme case-blind (it lower-cases before the allow-list check, and RFC 3986 3.1 makes schemes case-insensitive), but the file handling just below it tests `"file".equals(scheme)` twice, so a `FILE:` or `File:` scheme takes neither branch.

1. the empty-authority allowance is skipped, so a validator built for the file scheme rejects `FILE:///etc/hosts` while accepting `file:///etc/hosts`.
2. the drive-letter guard is skipped, so with `ALLOW_LOCAL_URLS` the authority `FILE://C:/some.file` is accepted although the byte-identical `file://C:/some.file` is rejected as it should be (`testValidator276` already pins the lower case form as never valid).

Computed the file-scheme test once with `equalsIgnoreCase` and reused it in both places. Left unfixed the second case is a validation bypass: an allow-list keyed on the scheme lets a crafted upper case authority slip past the guard the lower case form is stopped by. Added `testFileSchemeCaseInsensitive`, which fails on master and passes with the change.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
+++ b/src/main/java/org/apache/commons/validator/routines/UrlValidator.java
@@ -384,12 +384,15 @@ public class UrlValidator implements Serializable {
             return false;
         }
         final String authority = uri.getRawAuthority();
+        // URI schemes are case-insensitive (RFC 3986 3.1) and isValidScheme already matches case-blind,
+        // so the file: handling has to recognise the scheme case-blind too
+        final boolean fileScheme = "file".equalsIgnoreCase(scheme);
         // Special case - file: allows an empty authority, so only the authority check is skipped for it;
         // the path, query and fragment below are validated as they are for any other scheme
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hello @sahvx655-wq 

Thank you for the PR.

Here's a potential minor gap: the new test only checks FILE://C:/some.file for the drive-letter guard, not the mixed File://C:/…. 

The guard is case-blind so it would work, but an explicit mixed-case negative test would make the intent crystal clear.

Also, test for an empty-authority with mixed case and a path that would otherwise be rejected, like File:/C:/path/to/dir/.

garydgregory:
@sahvx655-wq Thank you, merged 🚀

PR #435, opened 2026-08-26
Reject under-length codes in SedolCheckDigit

sahvx655-wq:
`SedolCheckDigit.SEDOL_CHECK_DIGIT.isValid` is a public `CheckDigit` entry point with no wrapping validator, yet its `calculateModulus` override only rejects codes longer than the seven-character weight table. A shorter string whose weighted digits happen to sum to a multiple of ten therefore validates: `isValid("55")`, `isValid("550")`, `isValid("5500")` and `isValid("0055")` all return true, though a SEDOL is defined as exactly seven characters. I traced it back from the weight loop after noticing the base-class `isValid` does no length check of its own, so this one-sided `>` comparison is the only thing standing between a short input and a false positive.

The guard now also requires the validated form (the code passed with its check digit) to be exactly seven characters, while the calculate path still receives the six-character base and so keeps only the over-length check. Holding the length rule inside `calculateModulus`, next to the weight table it depends on, keeps the validate and calculate callers consistent without either repeating it. Left as it was, a caller relying on `SEDOL_CHECK_DIGIT` to screen out malformed identifiers would accept a two-character string as a valid security identifier.

### DIFF
--- a/src/main/java/org/apache/commons/validator/routines/checkdigit/SedolCheckDigit.java
+++ b/src/main/java/org/apache/commons/validator/routines/checkdigit/SedolCheckDigit.java
@@ -66,8 +66,13 @@ public final class SedolCheckDigit extends ModulusCheckDigit {
      */
     @Override
     protected int calculateModulus(final String code, final boolean includesCheckDigit) throws CheckDigitException {
-        if (code.length() > POSITION_WEIGHT.length) {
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you @sahvx655-wq , merged 🚀 
What does the spec about lengths of 6 and below? How do we do there?

